PR #7, opened 2015-01-16
U/pgee/dm 1013

jdswinbank:
PR for review.

### DIFF
--- a/python/lsst/meas/base/plugins.py
+++ b/python/lsst/meas/base/plugins.py
@@ -137,23 +137,38 @@ class SingleFrameClassificationPlugin(SingleFramePlugin):
         SingleFramePlugin.__init__(self, config, name, schema, metadata)
         self.keyProbability = schema.addField(name + "_value", type="D",
                                               doc="Set to 1 for extended sources, 0 for point sources.")
+        self.keyFlag = schema.addField(name + "_flag", type="Flag", doc="Set to 1 for any fatal failure.")
 
     def measure(self, measRecord, exposure):
         modelFlux = measRecord.getModelFlux()
         modelFluxErr = measRecord.getModelFluxErr()
+        modelFluxFlag = measRecord.getModelFluxFlag()
         psfFlux = measRecord.getPsfFlux()
         psfFluxErr = measRecord.getPsfFluxErr()
-        flux1 = self.config.fluxRatio*modelFlux + self.config.modelErrFactor*modelFluxErr
-        flux2 = psfFlux + self.config.psfErrFactor*psfFluxErr
-        if flux1 < flux2:
-            measRecord.set(self.keyProbability, 0.0)
+        psfFluxFlag = measRecord.getPsfFluxFlag()
+        flux1 = self.config.fluxRatio*modelFlux
+        if not self.config.modelErrFactor == 0:
+            flux1 += self.config.modelErrFactor*modelFluxErr
+        flux2 = psfFlux
+        if not self.config.psfErrFactor == 0:
+            flux2 += self.config.psfErrFactor*psfFluxErr
+
+        # A generic failure occurs when either FluxFlag is set to True
+        # A generic failure also occurs if either calculated flux value is NAN:
+        #     this can occur if the Flux field itself is NAN,
+        #     or the ErrFactor != 0 and the FluxErr is NAN
+        if numpy.isnan(flux1) or numpy.isnan(flux2) or modelFluxFlag or psfFluxFlag:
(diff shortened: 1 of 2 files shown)


python/lsst/meas/base/plugins.py
```diff
@@ -137,23 +137,39 @@ def __init__(self, config, name, schema, metadata):
         SingleFramePlugin.__init__(self, config, name, schema, metadata)
         self.keyProbability = schema.addField(name + "_value", type="D",
                                               doc="Set to 1 for extended sources, 0 for point sources.")
+        self.keyFlag = schema.addField(name + "_flag", type="Flag",
+                                              doc="Set to 1 for any fatal failure.")
```
jdswinbank: The indentation here is a bit weird -- it's not a fatal flaw, but it would be nicer if the "d" of "doc" appeared directly beneath the "n" of "name". In fact, the text is short enough that everything would fit on a single line, which would be better still.

python/lsst/meas/base/plugins.py
```diff
@@ -137,23 +137,39 @@ def __init__(self, config, name, schema, metadata):
...
-        flux2 = psfFlux + self.config.psfErrFactor*psfFluxErr
-        if flux1 < flux2:
-            measRecord.set(self.keyProbability, 0.0)
+        psfFluxFlag = measRecord.getPsfFluxFlag()
+        flux1 = self.config.fluxRatio*modelFlux
+        if not self.config.modelErrFactor == 0:
+            flux1 += self.config.modelErrFactor*modelFluxErr
+        flux2 = psfFlux
+        if not self.config.psfErrFactor == 0:
+            flux2 += self.config.psfErrFactor*psfFluxErr
```
jdswinbank: The logic here appears to have got more complicated for no very clear reason. If `self.config.modelErrFactor == 0`, then adding something multiplied by it has no effect (and the same for `self.config.psfErrFactor`, of course). What are the `if` statements adding?

tests/testClassification.py
```diff
@@ -36,7 +36,7 @@
 
 DATA_DIR = os.path.join(os.environ["MEAS_BASE_DIR"], "tests")
 
-class SFMTestCase(lsst.utils.tests.TestCase):
+class SFMTestCase (lsst.meas.base.tests.AlgorithmTestCase):
```
jdswinbank: There shouldn't be a space between the 'e' and the '('.

tests/testClassification.py
```diff
@@ -75,6 +75,116 @@ def testAlgorithm(self):
             else:
                 self.assertEqual(probability, 1.0)
 
+    def testFlags(self):
+        """ Test all the failure modes of this algorithm, as well as checking that it succeeds when it should
```
jdswinbank: Technically, the coding standards insist this line should end with a ".".

I think the indentation of this docstring looks strange -- I can't find anything in the standards which says it's forbidden, but it's inconsistent with the other Python docstrings (e.g. in `meas.base.plugins`). I think you should consider reformatting it.

tests/testClassification.py
```diff
@@ -75,6 +75,116 @@ def testAlgorithm(self):
...
+        task.plugins["base_ClassificationExtendedness"].measure(source, exp)
+        self.assertFalse(source.get("base_ClassificationExtendedness_flag"))
+
+        sfm_config.plugins["base_ClassificationExtendedness"].psfErrFactor = 1.
+        source = measCat.addNew()
+        source.set("base_PsfFlux_flux", 100)
+        source.set("base_SincFlux_fluxSigma", 1)
+        source.set("base_SincFlux_flux", 200)
+        task.plugins["base_ClassificationExtendedness"].measure(source, exp)
+        self.assertTrue(source.get("base_ClassificationExtendedness_flag"))
```
jdswinbank: You could write the same tests with a lot less code by defining a method which takes the source parameters and expected flag value as arguments and then calling it repeatedly.

python/lsst/meas/base/plugins.py
pgee2000: This code is basically coding around the possibility that the Err is a NAN.

tests/testClassification.py
pgee2000: I thought of that too, but it didn't seem like it mattered for a unit test.  And there are little differences from test to test.

python/lsst/meas/base/plugins.py
pgee2000: This and all the following minor formatting problems were fixed.  I won't add a comment for each one.

PR #9, opened 2015-02-19
Measurement transformation framework.

smonkewitz:
Measurement algorithm plugins may now specify a "transformation" which will
convert their raw outputs into a calibrated form, suitable for database
ingestion etc.

### DIFF
--- a/python/lsst/meas/base/base.py
+++ b/python/lsst/meas/base/base.py
@@ -34,7 +34,7 @@ import lsst.pex.config
 
 from .baseLib import *
 from .noiseReplacer import *
-from .transforms import NullTransform
+from .transforms import PassThroughTransform
 
 # Exceptions that the measurement tasks should always propagate up to their callers
 FATAL_EXCEPTIONS = (MemoryError, FatalAlgorithmError)
@@ -274,10 +274,10 @@ class BasePlugin(object):
         measurement quantities to calibrated units. Calibrated data is then
         provided in a separate output table.
 
-        By default, we assume that no transformation is appropriate, and
-        therefore return NullTransform.
+        By default, we copy everything from the input to the output without
+        transformation.
         """
-        return NullTransform
+        return PassThroughTransform
 
 
 class SourceSlotConfig(lsst.pex.config.Config):
--- a/python/lsst/meas/base/base.py
+++ b/python/lsst/meas/base/base.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 #
 # LSST Data Management System
-# Copyright 2008-2015 LSST Corporation.
+# Copyright 2008-2015 AURA/LSST.
 #
 # This product includes software developed by the
 # LSST Project (http://www.lsst.org/).
@@ -34,6 +34,7 @@ import lsst.pex.config
 
 from .baseLib import *
 from .noiseReplacer import *
+from .transforms import NullTransform
 
 # Exceptions that the measurement tasks should always propagate up to their callers
 FATAL_EXCEPTIONS = (MemoryError, FatalAlgorithmError)
@@ -261,6 +262,23 @@ class BasePlugin(object):
                    "please report this as a bug (the full traceback is above)."
                    % self.__class__.__name__)
         raise NotImplementedError(message)
+
+    @staticmethod
+    def getTransformClass():
+        """!
+        Get the measurement transformation appropriate to this plugin.
+
+        This returns a subclass of MeasurementTransform, which may be
+        instantiated with details of the algorithm configuration and then
+        called with information about calibration and WCS to convert from raw
+        measurement quantities to calibrated units. Calibrated data is then
+        provided in a separate output table.
+
+        By default, we assume that no transformation is appropriate, and
+        therefore return NullTransform.
+        """
+        return NullTransform
+
 
 class SourceSlotConfig(lsst.pex.config.Config):
     """!
--- a/python/lsst/meas/base/transforms.py
+++ b/python/lsst/meas/base/transforms.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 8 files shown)


python/lsst/meas/base/base.py
```diff
@@ -262,6 +263,23 @@ def fail(self, measRecord, error=None):
                    % self.__class__.__name__)
         raise NotImplementedError(message)
 
+    @staticmethod
+    def getTransformClass():
+        """!
+        Get the measurement transformation appropriate to this plugin.
+
+        This returns subclass of MeasurementTransform, which may be
```
smonkewitz: s/subclass/a subclass/

python/lsst/meas/base/base.py
```diff
@@ -262,6 +263,23 @@ def fail(self, measRecord, error=None):
...
+        instantiated with details of the algorithm configuration and then
+        called with information about calibration and WCS to convert from raw
+        measurement quantities to calibrated units. Calibrated data is then
+        provided in a separate output table.
+
+        By default, we assume that no transformation is appropriate, and
+        therefore return NullTransform.
+        """
```
smonkewitz: I think that PassThroughTransform might be a less surprising default - or is the expectation that by default you don't want to present users with uncalibrated data, and so all such should be removed prior to database ingest?

python/lsst/meas/base/transforms.py
```diff
@@ -0,0 +1,103 @@
...
+When a transformer is called, it is handed a `SourceCatalog` containing the
+measurements to be transformed, a `BaseCatalog` in which to store results, and
+information about the WCS and calibration of the data. It may be safely
+assumed that both are contiguous in memory, thus a ColumnView may be used for
+efficient processing.
+
+Transformations can be defined in Python or in C++. Python code should inherit
+from `MeasurementTransform`, following its interface. C++ code inherits from
+`BaseTransform`. `TransformWrapper` wraps the C++ implementation so that it
+provides a uniform interface with Python.
```
smonkewitz: I really like your documentation (and not just here).

python/lsst/meas/base/transforms.py
```diff
@@ -0,0 +1,103 @@
...
+    This is intended as the default for measurements for which no other
+    transformation is specified.
+    """
+    def __call__(self, inputCatalog, outputCatalog, wcs, calib):
+        pass
+
+
+class PassThroughTransform(MeasurementTransform):
+    """
+    Copy all fields named after the measurement plugin from input to output, wihtout transformation.
```
smonkewitz: s/wihtout/without/

python/lsst/meas/base/transforms.py
```diff
@@ -0,0 +1,103 @@
...
+    def __call__(self, inputCatalog, outputCatalog, wcs, calib):
+        pass
+
+
+class TransformWrapper(object):
+    """
+    Wrap a C++-style transformation.
+
+    C++ code requires a `Control`, rather than a `Config`, object. This class
+    converts the latter to the former when the transformer is instantiated.
```
smonkewitz: Do you mean invoked or called, rather than instantiated?

python/lsst/meas/base/wrappers.py
```diff
@@ -116,9 +118,15 @@ def wrapAlgorithm(Base, AlgClass, factory, executionOrder, name=None, Control=No
...
-                            getExecutionOrder=staticmethod(getExecutionOrder)))
+    if TransformClass:
+        PluginClass = type(AlgClass.__name__ + Base.__name__, (Base,),
+                           dict(AlgClass=AlgClass, ConfigClass=ConfigClass, factory=staticmethod(factory),
+                                getExecutionOrder=staticmethod(getExecutionOrder),
+                                getTransformClass=staticmethod(lambda: TransformClass)))
+    else:
+        PluginClass = type(AlgClass.__name__ + Base.__name__, (Base,),
+                           dict(AlgClass=AlgClass, ConfigClass=ConfigClass, factory=staticmethod(factory),
+                                getExecutionOrder=staticmethod(getExecutionOrder)))
```
smonkewitz: Can't you avoid stuttering the common initialization code by building the dict beforehand and adding a key for getTransformClass only if necessary?

tests/testTransform.py
```diff
@@ -0,0 +1,171 @@
...
+        # NB the choice of algorithm and executionOrder is arbitrary
+        singleFrame, forced = measBase.wrapSimpleAlgorithm(measBase.PsfFluxAlgorithm,
+                                                           Control=measBase.PsfFluxControl,
+                                                           executionOrder=0.0, doRegister=False)
+        self.assertEqual(singleFrame.getTransformClass(), measBase.BasePlugin.getTransformClass())
+        # Unless we override
+        singleFrame, forced = measBase.wrapSimpleAlgorithm(measBase.PsfFluxAlgorithm,
+                                                           Control=measBase.PsfFluxControl,
+                                                           executionOrder=0.0, doRegister=False,
+                                                           TransformClass=measBase.PassThroughTransform)
```
smonkewitz: Looks like there should be a check that singleFrame.getTransformClass() is PassThroughTransform here.

python/lsst/meas/base/base.py
jdswinbank: I have no strong opinion on this, but the current behaviour is what was requested by @TallJimbo; perhaps he'll pitch in here.

Jim -- if it's not obvious from the above, the question is: what should the calibration transform framework do when applied to a measurement plugin which does not have a specific transform specified? I think you previously indicated that it should not copy any of the measurements; the alternative Serge is suggesting is to copy all of them.

python/lsst/meas/base/transforms.py
jdswinbank: I actually _did_ mean instantiated here, but the way it's written is pretty confusing, so you're quite right to call me on it. I'll make it clearer.

python/lsst/meas/base/transforms.py
jdswinbank: Done.

python/lsst/meas/base/base.py
jdswinbank: Done.

tests/testTransform.py
jdswinbank: Done.

python/lsst/meas/base/wrappers.py
jdswinbank: Done.

python/lsst/meas/base/base.py
TallJimbo: I think Serge is probably right here.  There will be many columns (flags in particular, though we may want to handle them specially) that will not need to be transformed, but there will be very few that should be excluded from the calibrated catalog entirely.

jdswinbank: Done.

PR #11, opened 2015-03-16
meas_base Review for DM-420 

TallJimbo:


src/GaussianCentroid.cc
```diff
@@ -499,7 +498,6 @@ MAKE_TWODG(afw::image::Image<float>);
 MAKE_TWODG(afw::image::Image<double>);
 MAKE_TWODG(afw::image::Image<int>);
 
-} // end anonymous namespace
```
TallJimbo: I don't think there's any reason to remove the anonymous namespace here - I'm guessing this is a relic from when you had planned to make the {{twodg}} function itself public, rather than a wrapper (I like what you've done with the wrapper much better, but I think that means you can put the anonymous namespace back in).

src/GaussianCentroid.cc
```diff
@@ -519,6 +517,18 @@ GaussianCentroidAlgorithm::GaussianCentroidAlgorithm(
 }
 
 
+template<typename PixelT>
+afw::geom::Point2D GaussianCentroidAlgorithm::fitCentroid(afw::image::Image<PixelT> const& image,
+                  double x,
+                  double y
+) {
```
TallJimbo: Indentation for these function arguments should be:
```
afw::geom::Point2D GaussianCentroidAlgorithm::fitCentroid(
    afw::image::Image<PixelT> const& image,
    double x,
    double y
) {
```
(there are other allowed forms, too - see section 6-6 of the [style guide](https://confluence.lsstcorp.org/pages/viewpage.action?pageId=16908737) - but the above form is the one you're closest to already)

tests/sincPhotSums.py
```diff
@@ -0,0 +1,254 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+# -*- lsst-python -*-
+
+import lsst.pex.policy as pexPolicy
```
TallJimbo: This import isn't needed.

TallJimbo:
One general note: I see that there's been another merge to master since you last rebased, and it was a fairly major ticket.  You may want to rebase one more time before merging, just to make sure you can get any conflicts dealt with more easily.

PR #14, opened 2015-04-06
Review for DM-2306: Transforms for Centroid Measurements

TallJimbo:


include/lsst/meas/base/GaussianCentroid.h
```diff
@@ -127,6 +127,12 @@ class GaussianCentroidAlgorithm : public SimpleAlgorithm {
     SafeCentroidExtractor _centroidExtractor;
 };
 
+class GaussianCentroidTransform : public CentroidTransform {
+public:
+    typedef GaussianCentroidControl Control;
+    GaussianCentroidTransform(Control const & ctrl, std::string const & name, afw::table::SchemaMapper & mapper);
+};
```
TallJimbo: It's much more common in our codebase to have blank lines between each method, typedef, and member in a class (see standard 6-19, though reading that just now is the first time I've noticed that we actually suggest two blank lines in source files).

TallJimbo:
This looks really good - I especially like the `pythonprepend` trick you used to make C++ and Python objects callable the same way despite the Control/Config difference.

Aside from the really nitpicky things above, my only big-picture question is whether we need the {{CoordResult}} and {{CoordResultKey}} objects (or, for that matter, `MagResult` and `MagResultKey`).  Since measurement algorithms won't be producing those outputs directly, I think only the `Transform` objects will use them, and they just aggregate two quantities that already have `FunctorKeys`.  I'm inclined to say that we don't need them *if* we can move all the convenience functionality they provide into the `Transform` base classes themselves.  Quite happy to entertain arguments to the contrary, though.

jdswinbank:
Thanks for the review!

Regarding your 'big picture' comment, I guess this depends on the likely longer-term uses of the code. It seemed not implausible to me that sometime in the future folks would want to unpack celestial positions from tables into code, and this makes it easy for them to do so. You probably have a better idea than I do as to whether that's long-term-useful, though, and I certainly agree that it's not being used anywhere today, so I have no objection to rearranging things as you suggest.

TallJimbo:
I suspect anyone wanting extract these values from tables would also want to extract the flags as well, and hence aggregating 2 of the 3 sets of fields they'd want doesn't really gain them much (and unfortunately, extracting the flags in a generic-but-safe way is much harder).

jdswinbank:
@TallJimbo -- branch updated to take account of your suggestions. Currently waiting for buildbot; once that runs successfully, I'll merge to master unless you indicate you'd like to take a further look.

PR #15, opened 2015-04-09
Review for meas_base DM-2307: transformations for shape measurements

TallJimbo:


include/lsst/meas/base/ShapeUtilities.h
```diff
@@ -130,7 +130,8 @@ class ShapeResultKey : public afw::table::FunctorKey<ShapeResult>  {
         afw::table::Schema & schema,
         std::string const & name,
         std::string const & doc,
-        UncertaintyEnum uncertainty
+        UncertaintyEnum uncertainty,
+        afw::table::CoordinateType coordType=afw::table::CoordinateType::PIXEL
```
TallJimbo: You use this like an `enum class` here but like a regular `enum` elsewhere.  Unless I've caught you in the midst of changing it, one of these is wrong.

src/SdssShape.cc
```diff
@@ -882,5 +882,53 @@ INSTANTIATE_PIXEL(int);
...
+) const {
+    _fluxTransform(inputCatalog, outputCatalog, wcs, calib);
+    _centroidTransform(inputCatalog, outputCatalog, wcs, calib);
 
+    CentroidResultKey centroidKey(inputCatalog.getSchema()[_name]);
+    ShapeResultKey inShapeKey(inputCatalog.getSchema()[_name]);
+
+    afw::table::SourceCatalog::const_iterator inSrc = inputCatalog.begin();
+    afw::table::BaseCatalog::iterator outSrc = outputCatalog.begin();
+    for (; inSrc < inputCatalog.end() && outSrc < outputCatalog.end(); ++inSrc, ++outSrc) {
```
TallJimbo: Docs claim this throws `LengthError` if sizes don't match; this looks like you're just using the smaller of the two sizes in this case.  Maybe better to check for that and throw up front, then only check one iterator in the loop iteration.

jdswinbank: In fact, the call to `_fluxTransform()` at line 885 is where the exception will be thrown; it just gets propagated through this to the caller. That should be clearer though, and you're right that there's no point duplicating the check in the loop.

include/lsst/meas/base/ShapeUtilities.h
jdswinbank: In C++11, it's permitted to give scope to a regular `enum`. That is, `afw::table::CoordinateType::PIXEL` and `afw::table::PIXEL` are equivalent, and this does, in fact, work. I agree that this code is ugly & inconsistent, though.

It looks as though Swig is happy with `enum class`, so I think switching everything to that is a clear win.

PR #19, opened 2015-06-22
meas_base review for DM-1945: tasks for multi-band coadd processing

TallJimbo:


python/lsst/meas/base/forcedPhotCoadd.py
```diff
@@ -102,7 +105,7 @@ def fetchReferences(self, dataRef, exposure):
     def _makeArgumentParser(cls):
         parser = lsst.pipe.base.ArgumentParser(name=cls._DefaultName)
         parser.add_id_argument("--id", "deepCoadd_forced_src", help="data ID, with raw CCD keys + tract",
-                               ContainerClass=lsst.coadd.utils.CoaddDataIdContainer)
+                               ContainerClass=lsst.coadd.utils.ExistingCoaddDataIdContainer)
```
TallJimbo: I understand at least part of why @PaulPrice started using this (it makes it easier to deal with patches where there are no data), but I'm not totally sold on whether silently skipping data IDs where coaddition failed is the right logic; it seems like that could lead to problems being discovered later than they ought to be.

@laurenam, do you know if this change is necessary (rather than just a convenience) for the rest of this issue?  If not, I'd be inclined to suggest we take it off this issue and RFC it before we bring it over.  If so, I think we bring it over now but open an issue for reconsidering it in the future.

python/lsst/meas/base/references.py
```diff
@@ -150,10 +160,20 @@ class CoaddSrcReferencesTask(BaseReferencesTask):
...
     ConfigClass = CoaddSrcReferencesConfig
-
-    def getSchema(self, butler):
-        """Return the schema of the reference catalog"""
-        return butler.get(self.config.coaddName + "Coadd_src_schema", immediate=True).getSchema()
+    datasetSuffix = "src" # Suffix to add to "Coadd_" for dataset name
+
+    def __init__(self, butler=None, schema=None, **kwargs):
+        """! Initialize the task.
+        Additional keyword arguments (forwarded to BaseReferncesTask.__init__):
```
TallJimbo: Typo here: "BaseReferncesTask"

python/lsst/meas/base/references.py
```diff
@@ -168,7 +188,7 @@ def fetchInPatches(self, dataRef, patchList):
 
         The given dataRef must include the tract in its dataId.
         """
-        dataset = self.config.coaddName + "Coadd_src"
+        dataset = self.config.coaddName + "Coadd_" + self.datasetSuffix
```
TallJimbo: Once we get to three terms in a string concatenation, I think it's probably time to use e.g. `"{}Coadd_{}".format(self.config.coaddName, self.datasetSuffix)` (or `%`, but I'm not sure that's compatible with Python 3).

python/lsst/meas/base/forcedPhotCoadd.py
laurenam: I think it's fine to leave it as just CoaddDataIdContainer (I had asked you guys about this earlier and you leaned the other way, but I'm happy to revert!)

TallJimbo: Heh, I guess I must react to changes differently when my I have my reviewer hat on.

laurenam: @TallJimbo , did you want to me revert to CoaddDataIdContainer in multiBand.py as well (you didn't comment on it there, so I just want to make sure this was a global comment on this functionality).

TallJimbo: Yes, it was intended as a global comment (but if you didn't interpret it that way, that's fine too - I see you've already merged).

laurenam: Indeed I did go with the global interpretation, so the merge pipe_tasks has been reverted.  I did merge the addition of the class to coadd_utils, so if/when we every want to use it, it's there.

PR #20, opened 2015-06-26
meas_base review for DM-436: port aperture corrections from HSC

TallJimbo:


python/lsst/meas/base/base.py
```diff
@@ -39,6 +39,26 @@
...
+    _ApCorrNameSet.add(str(name))
+
+def getApCorrNameSet():
+    """!Return a copy of the set of names of algorithms that measure fluxes that can be aperture corrected
+
+    For example the returned set will likely include "slot_PsfFlux" and "slot_GaussianFlux"
+    """
+    global _ApCorrNameSet
+    return _ApCorrNameSet.copy()
+
```
TallJimbo: The names you're using as examples here (`slot_PsfFlux`, `slot_GaussianFlux`) aren't actually algorithm names - `slot_PsfFlux` is the aliase created to identify which algorithm is in the PsfFlux slot, and `slot_GaussianFlux` doesn't actually exist.  The names of the algorithms you're thinking of are `base_PsfFlux` and `base_GaussianFlux` (the first component of the name indicates the package the algorithm lives in; "base" is for "meas_base").

python/lsst/meas/base/base.py
```diff
@@ -105,14 +125,19 @@ def register(self, name, PluginClass):
         The name will be used as a prefix for all fields produced by the Plugin, and it
         should generally contain the name of the Plugin or Algorithm class itself
         as well as enough of the namespace to make it clear where to find the code.
+
+        Set canApCorr True if this algorithm measure a flux that can be aperture corrected;
+        the name will be added to the set retrieved by getApCorrNameSet.
         """
         lsst.pex.config.Registry.register(self, name, self.Configurable(name, PluginClass))
+        if canApCorr:
+            addApCorrName(name)
```
TallJimbo: While this mechanism for registering aperture-corrected names is what we want for most algorithms, it should be a convenience option that delegates to a more general mechanism that allows a single algorithm to register multiple flux fields to be corrected.  The CModel algorithm in `meas_modelfit` has three flux fields, for instance: `modelfit_CModel_flux`, `modelfit_CModel_exp_flux`, and `modelfit_CModel_dev_flux`.

The documentation should also state explicitly that any name added to the registry is the prefix for a trio of `{name}_flux`, `{name}_fluxSigma`, and `{name}_flag` fields.

python/lsst/meas/base/base.py
```diff
@@ -129,29 +154,29 @@ class ForcedTransformedCentroidPlugin(ForcedPlugin):
...
-        """Call each plugin in the map which has a measure() method
+        """Call each plugin in the map which has a .config.measure() method
         """
         for plugin in self.itervalues():
             if plugin.config.doMeasure:
                 yield plugin
 
     def iterN(self):
-        """Call each plugin in the map which has a measureN() method
+        """Call each plugin in the map which has a .config.measureN() method
```
TallJimbo: This change isn't correct: the config classes do not have `measure()` or `measureN()` methods, the plugin classes do.  This does highlight the fact that the old description isn't quite accurate: it only calls plugins for which those methods are present *and* we've enabled it in config (the config options are class attributes that are always False when the methods are not present).

python/lsst/meas/base/base.py
```diff
@@ -181,24 +206,28 @@ class BasePlugin(object):
...
-        Must be reimplemented (as a static or class method) by concrete derived classes.
+        CENTROID_ORDER  centroids and other algorithms that require only a Footprint and its Peaks as input
+        SHAPE_ORDER     shape measurements and other algorithms that require getCentroid() to return
+                        a good centroid (in addition to a Footprint and its Peaks).
+        FLUX_ORDER      flux algorithms that require both getShape() and getCentroid(),
+                        in addition to a Footprint and its Peaks
+        APCORR_ORDER    algorithms that operate on fluxes (e.g. aperture correction)
+        CLASSIFY_ORDER  algorithms that operate on aperture-corrected fluxes
+
+        Must be reimplemented as a class method by concrete derived classes.
```
TallJimbo: Let's just make the documentation for `APCORR_ORDER` "the executionOrder at which aperture correction is run"; we don't want to encourage plugins to use this order, I think, as we don't want them to have to worry about what happens in case of a tie.

python/lsst/meas/base/measureApCorr.py
```diff
@@ -0,0 +1,161 @@
...
+        @param[in] schema  catalog schema containing the flux field
+        """
+        self.flux = schema.find(name + "_flux").key
+        self.err = schema.find(name + "_fluxSigma").key
+        self.flag = schema.find(name + "_flag").key
+
+class MeasureApCorrConfig(lsst.pex.config.Config):
+    refFluxAlg = lsst.pex.config.Field(
+        doc = "Name of the measurement algorithm for the flux other measurements should be" +
+            "aperture corrected to match",
```
TallJimbo: Missing a space in the string concatenation here.

TallJimbo:
Just a general comment on nomenclature: `canApCorr` isn't really wrong, but I think `shouldApCorr` is a bit more appropriate.

python/lsst/meas/base/applyApCorr.py
```diff
@@ -0,0 +1,159 @@
...
+            three fields are used to generate keys:
+            - _algName_\_flux
+            - _algName_\_fluxSigma
+            - _algName_\_fluxFlag
+            three fields are added:
+            - _algName_\_ApCorr
+            - _algName_\_ApCorrSigma
+            - _algName_\_flag_apcorr
+        @param[in] algName  name of flux measurement algorithm, e.g. "base_PsfFlux"
+        """
```
TallJimbo: I find your format string syntax here a little unusual; does this get translated into boldface in HTML or something?

If that's all you're going for, I'd recommend using curly braces or something to set off the variable name (i.e. use Python `format()` syntax).

In general, I think if Doxygen markup interpretation requires us to do this much damage to the raw text form of a docstring (which is still visible in interactive help, after all), it's best to just disable the Doxygen markup by not including the initial `!`.

python/lsst/meas/base/applyApCorr.py
```diff
@@ -0,0 +1,159 @@
...
+                    oldFluxFlagState = source.get(apCorrInfo.fluxFlagKey)
+                    source.set(apCorrInfo.fluxFlagKey, True)
+
+                apCorr = 1.0
+                apCorrErr = 0.0
+                try:
+                    apCorr = apCorrModel.evaluate(center)
+                    apCorrErr = apCorrSigmaModel.evaluate(center)
+                except Exception:
+                    continue
```
TallJimbo: Would like a tighter exception specifier here (I realize that may be a problem you inherited from the HSC side).

python/lsst/meas/base/applyApCorr.py
```diff
@@ -0,0 +1,159 @@
...
+
+                flux = source.get(apCorrInfo.fluxKey)
+                fluxErr = source.get(apCorrInfo.fluxSigmaKey)
+                source.set(apCorrInfo.fluxKey, flux*apCorr)
+                a = flux/fluxErr
+                if apCorrErr == 0:
+                    b = float("inf")
+                else:
+                    b = apCorr/apCorrErr
+                source.set(apCorrInfo.fluxSigmaKey, abs(flux*apCorr)*math.sqrt(a*a + b*b))
```
TallJimbo: I realize this is what you inherited from the HSC side, but I think we want `apCorrErr==0` to lead to just `fluxSigma*=apCorr` (i.e. zero uncertainty in the aperture correction should mean we just scale the existing uncertainty by the correction factor).

By the way, this propagation of errors isn't actually correct - it includes some of the photon noise twice in the uncertainty, because it enters both through the aperture correction errors and the original flux errors.  But we don't actually have something better to change it to just yet (this is something that's still being figured out on the HSC side).  When we have that figured out, we'll probably change the formula entirely, but in the meantime we should get the zero-error case handled properly in case we decide it's actually better to set all the error to zero in an earlier stage.

TallJimbo: Actually, I just looked at the code on the HSC side, and the divide-by-zero problem is only coming up because there's an error in the formula: it should be:
```
a = fluxErr/flux
b = apCorrErr/apCorr
```

python/lsst/meas/base/applyApCorr.py
r-owen: It is markdown that puts "algName" in italics, indicating (to me, at least) that something is substituted for it. I agree it is ugly. In the old days I used <algName>_ApCorr, etc., but that made Doxygen unhappy, and I want Doxygen. If {algName}_ApCorr works then I'll use that; I agree it would be more readable in the source code.

python/lsst/meas/base/applyApCorr.py
r-owen: I have no idea what exceptions evaluate might raise when it cannot evalutate. Do you know?

python/lsst/meas/base/applyApCorr.py
r-owen: Many thanks. I'll clean this up.

python/lsst/meas/base/base.py
r-owen: How do you enable measure() or measureN in the config? I think it would be worth saying in the doc string. Right now I have this, which seems to beg the question:
{code}
        """Call each plugin in the map which has a measureN() method and is enabled in the config
        """
{code}

python/lsst/meas/base/applyApCorr.py
TallJimbo: I think `lsst.pex.exceptions.DomainError` is all you're likely to see (that could happen when trying to evaluate coadded aperture corrections at a position where there are no input images, which sounds like it shouldn't be possible, but experience on HSC is that it's surprisingly hard to get rid of all the ways that could accidentally happen).

python/lsst/meas/base/base.py
TallJimbo: It's just a matter of setting `config.doMeasure=True` or `config.doMeasureN=True` (for algorithms that don't support `measureN()`, `doMeasureN` is just a regular class attribute set to False, not a config field).

python/lsst/meas/base/applyApCorr.py
```diff
@@ -0,0 +1,156 @@
...
+        self.apCorrSigmaKey = schema.addField(
+            algName + "_ApCorr_Sigma",
+            doc = "aperture correction applied to %s" % (algName,),
+            type = float,
+        )
+        self.apCorrFlagKey = schema.addField(
+            algName + "_flag_apcorr",
+            doc = "set if unable to aperture correct %s" % (algName,),
+            type = "Flag",
+        )
```
TallJimbo: Capitalization and underscores are a bit inconsistent in these last three fields.  I propose:
 - `algName + "_apCorr"`
 - `algName + "_apCorrSigma"`  (note lack of underscore, to match e.g. `"_fluxSigma"`)
 - `algName + "_flag_apCorr"`

python/lsst/meas/base/applyApCorr.py
```diff
@@ -0,0 +1,156 @@
...
+
+class ApplyApCorrTask(lsst.pipe.base.Task):
+    """!Apply aperture corrections
+    """
+    ConfigClass = ApplyApCorrConfig
+    _DefaultName = "applyApCorr"
+
+    def __init__(self, schema, **kwds):
+        lsst.pipe.base.Task.__init__(self, **kwds)
+
```
TallJimbo: Newline here doesn't seem to be organizationally helpful.

python/lsst/meas/base/applyApCorr.py
```diff
@@ -0,0 +1,156 @@
...
+                # if a field in the registry is missing from the schema, silently ignore it.
+                pass
+
+    def run(self, catalog, apCorrMap):
+        """Apply aperture corrections to a catalog of sources
+
+        @param[in,out] catalog  catalog of sources
+        @param[in] apCorrMap  aperture correction map (an lsst.afw.image.ApCorrMap)
+        """
+
```
TallJimbo: Another newline I'd prefer to see go.

python/lsst/meas/base/applyApCorr.py
```diff
@@ -0,0 +1,156 @@
...
+        lsst.pipe.base.Task.__init__(self, **kwds)
+
+        self.apCorrInfoDict = dict()
+        apCorrNameSet = getApCorrNameSet()
+        ignoreSet = set(self.config.ignoreList)
+        missingNameSet = ignoreSet - set(apCorrNameSet)
+        if missingNameSet:
+            self.log.warn("Fields in ignoreList that are not in fluxCorrectList: %s" %
+                (sorted(list(missingNameSet)),))
+        for algName in apCorrNameSet - ignoreSet:
```
TallJimbo: Not a big deal, but I think it'd be appropriate to change the variable name here (and below) from `algName` to `apCorrName`.

python/lsst/meas/base/forcedMeasurement.py
```diff
@@ -298,16 +302,20 @@ def run(self, exposure, refCat, refWcs, idFactory=None, exposureId=None):
...
 
             # then process the parent record
             noiseReplacer.insertSource(refParentRecord.getId())
-            self.callMeasure(measParentRecord, exposure, refParentRecord, refWcs)
+            self.callMeasure(measParentRecord, exposure, refParentRecord, refWcs,
+                    beginOrder=beginOrder, endOrder=endOrder)
             self.callMeasureN(measParentCat[parentIdx:parentIdx+1], exposure,
-                              refParentCat[parentIdx:parentIdx+1])
+                              refParentCat[parentIdx:parentIdx+1],
+                    beginOrder=beginOrder, endOrder=endOrder)
```
TallJimbo: I'm okay with either indenting to the open-parentheses or indenting by a fixed amount, but it should be made consistent here.

python/lsst/meas/base/wrappers.py
```diff
@@ -127,6 +141,8 @@ def getExecutionOrder():
         if name is None:
             name = generateAlgorithmName(AlgClass)
         Base.registry.register(name, PluginClass)
+        if shouldApCorr:
+          addApCorrName(name)
```
TallJimbo: Indentation looks incorrect.

python/lsst/meas/base/applyApCorr.py
r-owen: I agree

python/lsst/meas/base/applyApCorr.py
r-owen: I like highlighting the call to the parent class

python/lsst/meas/base/applyApCorr.py
r-owen: Agreed. I tried to eliminate all use of "algorithmn name", replacing it with the clumsier but more accurate "flux field name prefix", but apparently had not fully succeeded. I changed "algName" to "name" as "apCorrName" did not seem to provide useful additional information in the contexts in which it was used.

python/lsst/meas/base/applyApCorr.py
r-owen: Agreed

python/lsst/meas/base/wrappers.py
r-owen: Good catch. Somehow my editor got into 2-space indent mode on that file.

jdswinbank:
This work was merged to master months ago (36e15925e1ffec559f5e32ec8b731d0afe9a3bdd); closing this PR to avoid confusion.

PR #21, opened 2015-07-23
meas_base review for tickets/DM-3182

TallJimbo:


TallJimbo:
I'm not convinced moving BaseMeasurementTask into its own file is an improvement, at least not on its own - before, we had a nice symmetry between base.py, forcedMeasurement.py, and sfm.py, where each held both the plugin base class and the task driver (and their configs), and that's now been lost.  Admittedly, base.py also had a bunch of other helper code, but if you're going to move BaseMeasurementTask into its own file, I think you should do the same with the other two files, and rename what's left over to e.g. basePlugin.py.  I'm ambivalent as to whether you do that or just leave the organization as it was, but I'd rather not land halfway in between.

python/lsst/meas/base/baseMeasurement.py
```diff
@@ -0,0 +1,258 @@
...
+        # remove it.
+        if self.config.slots.centroid is not None and self.plugins[self.config.slots.centroid] is None:
+            del self.plugins[self.config.slots.centroid]
+
+    def buildSubtasks(self):
+        """Build any subtasks.
+
+        self.schema must be available when this is called
+        """
+        self.makeSubtask("applyApCorr", schema=self.schema)
```
TallJimbo: I don't think this provides enough convenience/code-sharing to be worthwhile, but I don't care deeply enough about it to fight it if you feel strongly about it.

TallJimbo:
On the question of whether to just extract the `ApCorrMap` from the `Exposure`, I think that's a bigger issue than we want to try to address here, but maybe we can hedge a bit and provide both: have an `apCorrMap` argument that, if `None`, will be set to `Exposure.getApCorrMap()`.

python/lsst/meas/base/baseMeasurement.py
r-owen: It is safety I'm worried about, but I'll take it out for now. We can always add it later if we need it.

r-owen:
Regarding adding baseMeasurment: I agree about the lost symmetry. I will move the two base plugin classes into baseMeasurement.py, which restores symmetry and leaves base.py for really basic stuff and avoids a circular import, which was the primary motivation for the reorganization. I think renaming sfm.py to singleFrameMeasurement.py would also be an improvement for readability.

r-owen:
I am very reluctant to add apCorrMap as an argument to the measure methods until we demonstrate a need for it. I had started out using it but I feel it results in clutter and needless complexity. It will be very easy to add later if it proves useful.

TallJimbo:
Please go ahead and rename sfm.py if you're up to it.  It sounds like the right place to do it.  Maybe we should reconsider renaming base.py, too, if it's just the utility code now?

Leaving out the apCorrMap argument is fine with me.

PR #22, opened 2015-08-06
Review for DM-3400 (Eliminate circular aliases in slot centroid definition)

jdswinbank:


tests/centroid.py
```diff
@@ -137,11 +137,7 @@ class SingleFrameMeasurementTaskTestCase(utilsTests.TestCase):
     def mySetup(self, runCentroider=True):
         msConfig = measBase.SingleFrameMeasurementConfig()
         msConfig.algorithms.names = ["base_SdssCentroid"]
-        if not runCentroider:
-            msConfig.slots.centroid = None
-        else:
-            msConfig.slots.centroid = "base_SdssCentroid"
-
+        msConfig.slots.centroid = "base_SdssCentroid"
```
jdswinbank: You've eliminated the only use of the `runCentroider` argument. Drop it entirely?

TallJimbo: Oops, of course.

PR #23, opened 2015-08-13
meas_base DM-1954: HeavyFootprints in forced coadd photometry

TallJimbo:


python/lsst/meas/base/forcedMeasurement.py
```diff
@@ -250,45 +250,35 @@ def run(self, exposure, refCat, refWcs, idFactory=None, exposureId=None, beginOr
...
-        refList = list()
+        # First check that the reference catalog does not contain any children for which
+        # any member of their parent chain is not within the list.  This can occur at
+        # boundaries when the parent is outside and one of the children is within.
+        # Currently, the parent chain is always only one deep, but just in case, this
+        # code checks for any case where when the parent chain to a child's topmost
+        # parent is broken and raises an exception if it occurs.
+        #
+        # NOTE: this culling should to be accounted for at the reference catalog creation
+        # stage (of which there are several) but is not currently implemented.
```
TallJimbo: Trivial grammar error: "should to be accounted for".

TallJimbo: Also, instead of saying the culling is not implemented at the reference catalog stage, I think it's probably more accurate to say the code merely checks that this precondition is satisfied by whatever reference catalog provider is being paired with this code.  We do in fact have reference catalog creation code that guarantees these conditions are satisfied under the conditions in which they are run.

python/lsst/meas/base/forcedMeasurement.py
```diff
@@ -335,26 +325,18 @@ def run(self, exposure, refCat, refWcs, idFactory=None, exposureId=None, beginOr
...
 
-        This generate a new blank SourceRecord for each record in refCat, copying any
-        fields in ForcedMeasurementConfig.copyColumns.  This also transforms the Footprints
-        in refCat to the measurement coordinate system if it differs from refWcs, and attaches
-        these to the new SourceRecords.  Note that we do not currently have the ability to
-        transform heavy footprints, so when the reference and measure WCSs are different,
-        HeavyFootprints will be converted to regular Footprints, which makes it impossible
-        to properly measure blended objects.
+        This generates a new blank SourceRecord for each record in refCat.  Note that footprints
+        are attached to the sources using the class's implementation of attachFootprints().
```
TallJimbo: This last sentence now seems to be inaccurate: better to say that this method does *not* attach any `Footprints`, and doing so is up to the caller (who can call `attachedTransformedFootprints` if they don't have anything better).

python/lsst/meas/base/forcedMeasurement.py
```diff
@@ -368,12 +350,21 @@ def generateSources(self, exposure, refCat, refWcs, idFactory=None):
...
+        """!Default implementation for attaching Footprints to blank sources prior to measurement
+
+        Footprints for forced photometry must be in the pixel coordinate system of the image being
+        measured, while the actual detections may start out in a different coordinate system.
+        This default implementation transforms the Footprints from the reference catalog from the
+        refWcs to the exposure's Wcs, which downgrades HeavyFootprints into regular Footprints,
+        destroying deblend information.
+
+        Note that ForcedPhotImageTask inherits this method and all its derived classes must
+        implement this method to define how their Footprints should be generated.
```
TallJimbo: It's not actually "inherits": `ForcedPhotImageTask` _delegates_ to this method in its own `attachFootprints` method, so it's subclasses will use this implementation unless they override.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -130,8 +130,21 @@ def fetchReferences(self, dataRef, exposure):
...
+                if record.getParent() != 0:
+                    self.log.warn("Skipping reference %s (child of %s) with bad Footprint" %
+                    (record.getId(), record.getParent()))
+                else:
+                    self.log.warn("Skipping reference parent %s with bad Footprint" % (record.getId(),))
+                    badParents.add(record.getId())
+            elif record.getParent() not in badParents:
+                references.append(record)
+        references.sort()   # need to ensure catalog is in ID order so find methods work
+        return references
```
TallJimbo: With this change, the documentation at the top of this method is a bit imprecise: we're not returning just an iterable, we're returning a `SourceCatalog`.  And it's probably a good idea to mention in those docs that it's sorted by ID and guarantees that all included children have their parent included and all `Footprints` are valid.

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -65,48 +69,65 @@ def setDefaults(self):
...
 
     def __init__(self, butler=None, refSchema=None, **kwds):
+        """Initialize the task.
+
+        ForcedPhotImageTask takes two keyword arguments beyond the usual CmdLineTask arguments:
+         - schema: the Schema of the reference catalog, passed to the constructor of the references subtask
+         - butler: a butler that will be passed to the references subtask to allow it to load its Schema
+           from disk.
+        At least one of these arguments must be present; if both are, schema takes precedence.
+        """
```
TallJimbo: Actual function argument is `refSchema`, not `schema`; documentation should be changed to reflect this.

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -65,48 +69,65 @@ def setDefaults(self):
...
-                forced measurement on the exposure.
+        """!Measure a single exposure using forced detection for a reference catalog.
+
+        @param[in]  dataRef   An lsst.daf.persistence.ButlerDataRef. It is passed to the
+                              references subtask to obtain the reference WCS, the getExposure()
+                              method (implemented by derived classes) to read the measurement
+                              image, and the fetchReferences() method (implemented by derived
+                              classes) to get the exposure and load the reference catalog (see
+                              the CoaddSrcReferencesTask for more information).  Sources are
+                              generated wtih generateSources() in the measurement subtask.  These
```
TallJimbo: typo: "wtih"

python/lsst/meas/base/references.py
```diff
@@ -164,7 +177,7 @@ class CoaddSrcReferencesTask(BaseReferencesTask):
 
     def __init__(self, butler=None, schema=None, **kwargs):
         """! Initialize the task.
-        Additional keyword arguments (forwarded to BaseReferencesTask.__init__):
+        Additional keyword arguments (forwarded to BaseReferncesTask.__init__):
```
TallJimbo: typo: `BaseReferncesTask`

python/lsst/meas/base/references.py
```diff
@@ -183,8 +197,8 @@ def getWcs(self, dataRef):
 
     def fetchInPatches(self, dataRef, patchList):
         """!
-        An implementation of BaseReferencesTask.fetchInPatches that loads 'coadd_src' catalogs
-        using the butler.
+        An implementation of BaseReferencesTask.fetchInPatches that loads 'coadd_datasetSuffix'
+        catalogs using the butler.
```
TallJimbo: Maybe set off the "datasetSuffix" string somehow, so it's clear it's a variable, not a literal?

PR #24, opened 2015-08-18
Review for DM-3483: don't throw when transforming negative fluxes

TallJimbo:
When transforming data to calibrated form for database ingest we should not
allow the process to be derailed by a negative flux measurement.

### DIFF
--- a/include/lsst/meas/base/FluxUtilities.h
+++ b/include/lsst/meas/base/FluxUtilities.h
@@ -209,6 +209,21 @@ private:
     MagResultKey _magKey;
 };
 
+/**
+ *  Temporarily replace negative fluxes with NaNs
+ *
+ *  Instantiating a NoThrowOnNegativeFluxContext object will cause afw::image::Calib
+ *  objects to return NaN, rather than throwing, when asked to convert a negative flux to
+ *  a magnitude for the lifetime of the NoThrowOnNegativeFluxContext.
+ */
+class NoThrowOnNegativeFluxContext {
+public:
+    NoThrowOnNegativeFluxContext();
+    ~NoThrowOnNegativeFluxContext();
+private:
+    bool _throwOnNegative;
+};
+
 }}} // lsst::meas::base
 
 #endif // !LSST_MEAS_BASE_FluxUtilities_h_INCLUDED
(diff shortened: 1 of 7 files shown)


include/lsst/meas/base/FluxUtilities.h
```diff
@@ -209,6 +209,21 @@ class FluxTransform : public BaseTransform {
...
+ *  a magnitude for the lifetime of the NoThrowOnNegativeFluxContext.
+ */
+class NoThrowOnNegativeFluxContext {
+public:
+    NoThrowOnNegativeFluxContext();
+    ~NoThrowOnNegativeFluxContext();
+private:
+    bool _throwOnNegative;
+};
+
```
TallJimbo: I like this a lot.  I'd like it even better if we could integrate it more closely with Calib, but I don't think that's appropriate for this ticket (it's more disruptive), so I've opened DM-3489 for that.

src/ApertureFlux.cc
```diff
@@ -354,10 +354,15 @@ void ApertureFluxTransform::operator()(
...
     afw::table::BaseCatalog::iterator outSrc = outputCatalog.begin();
-    for (; inSrc < inputCatalog.end() && outSrc < outputCatalog.end(); ++inSrc, ++outSrc) {
-        for (std::size_t i = 0; i < _ctrl.radii.size(); ++i) {
-            FluxResult fluxResult = fluxKeys[i].get(*inSrc);
-            _magKeys[i].set(*outSrc, calib.getMagnitude(fluxResult.flux, fluxResult.fluxSigma));
+    {
+        // While noThrow is in scope, converting a negative flux to a magnitude
+        // returns NaN rather than throwing.
+        NoThrowOnNegativeFluxContext noThrow;
+        for (; inSrc < inputCatalog.end() && outSrc < outputCatalog.end(); ++inSrc, ++outSrc) {
```
TallJimbo: This part of these lines isn't what's been changed, but...
 - I'm not sure using `<` on iterators is safe in general; it's certainly more idiomatic to use `!=` on `for` loop exit conditions.
 - Iterator declarations seem a good place to use `auto`.

TallJimbo: Ah, to clarify, `<` is valid for RandomAccessIterators (which these are), but not for the more general `InputIterator` concept, so it's just bit more generic to use `!=`.

PR #26, opened 2015-10-21
Review for DM-2195 (Port safe coadd clipping from HSC)

jdswinbank:


include/lsst/meas/base/PixelFlags.h
```diff
@@ -33,7 +33,6 @@
 #include "lsst/afw/image/Exposure.h"
 #include "lsst/meas/base/Algorithm.h"
 #include "lsst/meas/base/FlagHandler.h"
```
jdswinbank: `FlagHandler.h` is now unused.

include/lsst/meas/base/PixelFlags.h
```diff
@@ -42,13 +41,20 @@ namespace lsst { namespace meas { namespace base {
  */
 class PixelFlagsControl {
 public:
-
+    LSST_CONTROL_FIELD(center, std::vector<std::string>,
+                       "List of mask planes for which to search center of footprint");
+    LSST_CONTROL_FIELD(any, std::vector<std::string>,
+                       "List of mask planes for which to search entire footprint");
```
jdswinbank: I find those docstrings a bit... lacklustre. Unless you already know in detail how the algorithm works, you wouldn't be any the wiser from reading them. Can we be any more descriptive?

(Actually, this whole algorithm needs proper documentation, but I think that's outside the scope of this ticket.)

Also, the names `any` and `center` are a bit too generic (below when we start referring to `control.any` it's pretty opaque).

include/lsst/meas/base/PixelFlags.h
```diff
@@ -42,13 +41,20 @@ namespace lsst { namespace meas { namespace base {
...
     /**
      *  @brief Default constructor
      *
      *  All control classes should define a default constructor that sets all fields to their default values.
```
jdswinbank: So why aren't the default values for the two fields defined above being set here?

include/lsst/meas/base/PixelFlags.h
```diff
@@ -42,13 +41,20 @@ namespace lsst { namespace meas { namespace base {
...
      */
-    PixelFlagsControl() {}
+    // We define this string to be consistent with the port from HSC. It also
+    // provideds a single place to change the string if a change is needed.
+    const std::string fname;
```
jdswinbank: I don't think this is necessary. Elsewhere, we hard-code flag fields to be called `flag`, and I'm not sure we gain anything by not doing so here.

include/lsst/meas/base/PixelFlags.h
```diff
@@ -74,8 +80,18 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
         CR,
         CR_CENTER,
         BAD,
+        CLIPPED,
```
jdswinbank: Given that you're not using `FlagHandler`, we can remove this enum entirely rather than add to it.

include/lsst/meas/base/PixelFlags.h
```diff
@@ -74,8 +80,18 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
...
         N_FLAGS
     };
+    
+    afw::table::Key<afw::table::Flag> _generalFailure;
+    afw::table::Key<afw::table::Flag> _offImageKey;
+    typedef std::map<std::string, afw::table::Key<afw::table::Flag> > KeyMap;
```
jdswinbank: Since we're in C++11, you can drop the space in `> >`.

include/lsst/meas/base/PixelFlags.h
```diff
@@ -74,8 +80,18 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
...
+    KeyMap _predefinedAnyKeys;
+    KeyMap _predefinedCenterKeys;
+    KeyMap _centerKeys;
+    KeyMap _anyKeys;
+
```
jdswinbank: Do these all need to be public?

Also, is there any benefit to keeping separate maps for the predefined keys and those set in the config?

include/lsst/meas/base/PixelFlags.h
```diff
@@ -94,10 +110,12 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
     ) const;
 
 private:
-
+    template <typename MaskedImageT>
+    class FootprintBits;
+    typedef afw::image::MaskedImage<float> MaskedImageT;
+    void updateFlags(const KeyMap *,const FootprintBits<MaskedImageT> *,afw::table::SourceRecord & ) const;
```
jdswinbank: `const` should come after the type name (coding standard 5-10).

Why is this taking pointers rather than references?

And does it need to be a member of `PixelFlagsAlgorithm`? Can we make it a free function in an anonymous namespace instead?

include/lsst/meas/base/PixelFlags.h
```diff
@@ -94,10 +110,12 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
...
     Control _ctrl;
     FlagHandler _flagHandler;
```
jdswinbank: We're no longer using `FlagHandler`.

src/PixelFlags.cc
```diff
@@ -28,11 +28,13 @@
 #include "lsst/afw/table/Source.h"
 #include "lsst/meas/base/PixelFlags.h"
 
+#include <cctype> // ::toupper
```
jdswinbank: ... `::tolower`.

src/PixelFlags.cc
```diff
@@ -28,11 +28,13 @@
...
+#include <algorithm> // std::transform
+
 namespace lsst { namespace meas { namespace base {
-namespace {
 
 template <typename MaskedImageT>
-class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
+class PixelFlagsAlgorithm::FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
```
jdswinbank: I don't understand why you've brought this out of the anonymous namespace.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
-        {"flag_interpolatedCenter", "Interpolated pixel in the source center"},
-        {"flag_saturated", "Saturated pixel in the source footprint"},
-        {"flag_saturatedCenter", "Saturated pixel in the source center"},
-        {"flag_cr", "Cosmic ray in the source footprint"},
-        {"flag_crCenter", "Cosmic ray in the source center"},
-        {"flag_bad", "Bad pixel in the source footprint"}
-    }};
-    _flagHandler = FlagHandler::addFields(schema, name, flagDefs.begin(), flagDefs.end());
+    // Add generic keys first, which don't correspond to specific mask planes
+    _generalFailure = schema.addField<afw::table::Flag>(name+"_"+_ctrl.fname,
```
jdswinbank: Throughout, please add spaces around your `+` operators.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+                                        "general failure flag, set if anything went wring");
+    _offImageKey = schema.addField<afw::table::Flag>(name+"_"+_ctrl.fname+"_offimage",
```
jdswinbank: Off image seems to have appeared here without any explanation -- it wasn't in the previous version of this code, and I don't think any of the commits refer to it.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+                                        "source's footprint includes suspect pixels");
+    // Flags that correspond to mask bits which occur in the center of the object
+    _predefinedCenterKeys["INTRP"] = schema.addField<afw::table::Flag>(name+"_"+_ctrl.fname+"_interpolatedCenter",
+                                        "Interpolated pixel in the source center");
+    _predefinedCenterKeys["SAT"] = schema.addField<afw::table::Flag>(name+"_"+_ctrl.fname+"_saturatedCenter",
+                                        "Saturated pixel in the source center");
+    _predefinedCenterKeys["CR"] = schema.addField<afw::table::Flag>(name+"_"+_ctrl.fname+"_crCenter",
+                                        "Cosmic ray in the source center");
+    _predefinedCenterKeys["SUSPECT"] = schema.addField<afw::table::Flag>(name+"_"+_ctrl.fname+"_suspectCenter",
+                                        "source's center is close to suspect pixels");
```
jdswinbank: Please capitalize consistently (should be "Source").

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+
+    // Read in the flags passed from the configuration, and add them to the schema
+    for (std::vector<std::string>::const_iterator i = _ctrl.center.begin(); i != _ctrl.center.end(); ++i){
```
jdswinbank: C++11 is your friend: `for (auto const & i: _ctrl.center) {` (and then note that `i` is a reference, not an iterator).

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+        std::string tname(*i);
```
jdswinbank: I realise you're using `tname` to mean "transformed name" here. But you already have a `name` variable... which is not the one you're transforming. Can you find something more descriptive?

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+                                        "source center is close to "+ *i + " pixels");
+    }
+    for (std::vector<std::string>::const_iterator i = _ctrl.any.begin(); i != _ctrl.any.end(); ++i){
+        std::string tname(*i);
+        std::transform(tname.begin(), tname.end(), tname.begin(), ::tolower);
+        _anyKeys[*i] = schema.addField<afw::table::Flag>(name+"_"+_ctrl.fname + "_" + tname,
+                                        "source footprint includes " + *i + " pixels");
+    }
 }
+void PixelFlagsAlgorithm::updateFlags(const KeyMap * maskFlagToPixelFlag, const FootprintBits<MaskedImageT> * func,
```
jdswinbank: `const` comes after the type name.

As per earlier comments, this looks like it should be a free function taking references rather than a member taking pointers.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+                                      afw::table::SourceRecord & measRecord) const{
+        lsst::afw::image::MaskPixel bitmask = 0x0;
+        for (KeyMap::const_iterator i = maskFlagToPixelFlag->begin(); i != maskFlagToPixelFlag->end(); ++i){
+            try {
+                bitmask = MaskedImageT::Mask::getPlaneBitMask(i->first);
+            } catch (pex::exceptions::InvalidParameterError & err){
+                throw LSST_EXCEPT(FatalAlgorithmError, err.what());
+            }
+            if (func->getBits() & MaskedImageT::Mask::getPlaneBitMask(i->first)){
+                measRecord.set(i->second, true);
```
jdswinbank: `bitmask` is never used in this function. Can you write:
```c++
try {
    if (func->getBits() & MaskedImageT::Mask::getPlaneBitMask(i->first)){
        measRecord.set(i->second, true);
    } catch catch (pex::exceptions::InvalidParameterError & err){
        throw LSST_EXCEPT(FatalAlgorithmError, err.what());
    }
}
```
?

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
-
     FootprintBits<MaskedImageT> func(mimage);
-
-//  Catch NAN in centroid estimate
-    if (lsst::utils::isnan(center.getX()) || lsst::utils::isnan(center.getY())) {
-        throw LSST_EXCEPT(pex::exceptions::InvalidParameterError,
-                          "Center point passed to PixelFlagsALgorithm is NaN");
+    
+    // Check if the measRecord has a valid centroid key, i.e. it was
+    // centroided
```
jdswinbank: Why the newline in the middle of the comment?

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+    afw::geom::Point2D center;
+    if (measRecord.getTable()->getCentroidKey().isValid()) {
+        center = measRecord.getCentroid(); 
+        //  Catch NAN in centroid estimate
+        if (lsst::utils::isnan(center.getX()) || lsst::utils::isnan(center.getY())) {
+            throw LSST_EXCEPT(pex::exceptions::RuntimeError,"Center point passed to PixelFlagsAlgorithm is Nan");
```
jdswinbank: Line is too long (110 characters, please).

Also, "NaN", not "Nan"!

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+        // be determined though the peak pixel
+        PTR(afw::detection::Footprint) footprint = measRecord.getFootprint();
+        // If there is no footprint or the footprint contains no peaks, throw
+        // a runtime error.
+        if (!footprint || footprint->getPeaks().empty()){
+            throw LSST_EXCEPT(pex::exceptions::RuntimeError,"No footprint, or no footprint peaks detected");
+        }
+        else {
+            center.setX(footprint->getPeaks().front().getFx());
+            center.setY(footprint->getPeaks().front().getFy());
```
jdswinbank: This is using _a_ peak, not necessarily _the_ peak, right? Is the first one special (ie, is the list sorted?), or is it a basically arbitrary choice? Please clarify in the comment above.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+        }
```
jdswinbank: In `SafeCentroid Extractor`, we have:
```c++
126         if (!_isCentroider) {
127             // set the general flag, because using the Peak might affect the current measurement
128             flags.setValue(record, FlagHandler::FAILURE, true);
129         }
```
That flag is no longer being set here. Is that change intentional?

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
     }
-//  Catch centroids off the image
+
+    //  Initialize bitmask to use in try statments
+    lsst::afw::image::MaskPixel bitmask = 0x0;
```
jdswinbank: Does it need to be initialized rather than just declared?

...in fact, do you need it at all?

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+
+    //  Catch centroids off the image
     if (!mimage.getBBox().contains(afw::geom::Point2I(center))) {
-       _flagHandler.setValue(measRecord, EDGE, true);
+        measRecord.set(_offImageKey, true);
+        auto iter = _predefinedAnyKeys.find("EDGE");
+        assert(iter != _predefinedAnyKeys.end());
+        measRecord.set(iter->second, true);
```
jdswinbank: The last three lines could be shortened to just `measRecord.set(_predefinedAnyKeys.at("EDGE"), true);`.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
     }
-    if (func.getBits() & MaskedImageT::Mask::getPlaneBitMask("CR")) {
-        _flagHandler.setValue(measRecord, CR, true);
+
+    // Set the EDGE flag if the bitmask has NO_DATA set
+    if (func.getBits() & MaskedImageT::Mask::getPlaneBitMask("NO_DATA")){
+        auto iter = _predefinedAnyKeys.find("EDGE");
+        assert(iter != _predefinedAnyKeys.end());
+        measRecord.set(iter->second, true);
     }
```
jdswinbank: As earlier, do you need the `try... catch` followed by the `if`? It seems like you could just surround the latter in the exception handler, and -- in the process -- eliminate `bitmask` entirely.

Also, again, these three lines could become one.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
 
+    // update the source record for the predefined any keys
+    updateFlags(&_predefinedAnyKeys,&func,measRecord);
```
jdswinbank: Use a space between function arguments (after the `,`s).

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+
     // Check for bits set in the 3x3 box around the center
-    afw::geom::Point2I llc(afw::image::positionToIndex(center.getX())-1, afw::image::positionToIndex(center.getY()) - 1);
+    afw::geom::Point2I llc(afw::image::positionToIndex(center.getX())-1,
+                           afw::image::positionToIndex(center.getY()) - 1);
```
jdswinbank: Using consistent spacing.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+    updateFlags(&_predefinedCenterKeys,&func,measRecord);
+
+    // update all the flags that are defined in the config
+    updateFlags(&_anyKeys, &func, measRecord);
+    updateFlags(&_centerKeys, &func, measRecord);
 }
 
 void PixelFlagsAlgorithm::fail(afw::table::SourceRecord & measRecord, MeasurementError * error) const {
-    _flagHandler.handleFailure(measRecord, error);
+    measRecord.set(_generalFailure,true);
```
jdswinbank: Spacing.

src/PixelFlags.cc
```diff
@@ -57,87 +59,153 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+    assert (error == NULL);
```
jdswinbank: Because the above is guaranteed never to throw a `MeasurementError`? I think it's worth a comment to explain.

tests/measureSources.py
```diff
@@ -237,15 +237,21 @@ def testPixelFlags(self):
         edge = mask.getPlaneBitMask('EDGE')
         bad = mask.getPlaneBitMask('BAD')
         nodata = mask.getPlaneBitMask('NO_DATA')
+        # Add in the non default mask plane CLIPPED (DM-2915)
```
jdswinbank: No need to sprinkle the ticket number through the comments -- it's easy enough to find with `git blame`.

tests/measureSources.py
```diff
@@ -294,6 +302,31 @@ def testPixelFlags(self):
             source,
             exp,
         )
+        # Test that if there is no center and centroider that the object should look at the footprint
+        plugin, cat = makePluginAndCat(measBase.PixelFlagsAlgorithm, "test", control)
```
jdswinbank: In fact there are three tests in quick succession here, and the first one tests what happens if there's no `Footprint`! 
Since you're following a very similar pattern repeatedly (make record, attempt to measure, check for failure), I wonder if this should actually be a series of separate tests calling a common implementation?

Also: a few lines above there's a comment (which GitHub won't let me annotate, because it's not been changed) to the effect that "the new code in meas_base InputUtilities.py now throws...". Is this code still testing `InputUtilities`, or is it testing your new code above? Please clarify the comment.

tests/measureSources.py
```diff
@@ -294,6 +302,31 @@ def testPixelFlags(self):
...
+        # The first test should raise exception because there is no footprint
+        source = cat.makeRecord()
+        self.assertRaises(
+                lsst.pex.exceptions.RuntimeError,
+                plugin.measure,
+                source,
+                exp,
+        )
+        # The second test will raise an error because no peaks are present
+        source.setFootprint(afwDetection.Footprint(afwGeom.Point2I(afwGeom.Point2D(x+x0, y+y0)),5))
```
jdswinbank: Spacing around operators.

Also, it seems like a bad idea to rely on `x` being what you left it at the end of the loop above. Next time somebody extends that loop, these tests will spontaneously break.

jdswinbank:
In addition to the line by line comments I added above, there's a more significant point here --

I understand why you have replaced the `FlagHandler`/`SafeCentroidExtractor` implementation with something more flexible, but I don't like it. We're introducing a bunch of code duplication which makes results much harder to predict (as I flagged above, I think your centroid implementation will set different flags in some circumstances than the `SafeCentroidExtractor`, and I don't know if that's intentional) and the whole infrastructure harder to maintain in the long run. In the short term, we've agreed that accumulating some level of technical debt while catching up with HSC is ok, so I don't think this need block us for now, but I have a couple of requests before this gets merged:

* Please file a technical debt ticket indicating that this is unfortunate and we need to think about it again in the future;
* Please add more explanation in commit messages, code comments and (optionally) Doxygen documentation describing and justifying the new implementation you've come up with here.

On a more minor level, please review the commit messages and make sure they make sense. In particular, the summary of aace6f is "Port of HSC-1166 safe coadd clipping", which is pretty misleading -- it's one small part of that port. It might also be nice to include reference to the specific HSC commits (by SHA1) which you're pulling over. And, being picky: please use consistent markup for adding your own commentary (I'm thinking of `~~~~ UPDATE~~~~` vs `~~Note~~`).

Finally, while reviewing this, I re-wrote some of the code to sanity check my comments. It's not a complete, polished up implementation, but it does include many of the changes I've requested above. Give me a shout if you'd like a copy to work from.

include/lsst/meas/base/PixelFlags.h
natelust: Let me think about this further, but the first reason that comes to mind, is that as of now, all the predefined keys are expected to be present in so many places in the code they have to be there, if they were in the same config structure, it may be possible for people to remove needed keys, breaking the stack. I agree ideally that we would have there be no explicit dependence on specific keys, but that would need to wait until the full flaghandler refactoring

include/lsst/meas/base/PixelFlags.h
RobertLuptonTheGood: I for one prefer to use pointers rather than non-const references.  It's more obvious to the reader that the function is modifying the values.

jdswinbank: Hmm, I'm not sure if I agree with the general point, but I might if you shouted loud enough.

In this case, though, const references are fine -- only the `SourceRecord` (which is already a non-const reference) is being modified.

jdswinbank: (Actually, given coding standard 5-24, const references SHOULD be used here.)

RobertLuptonTheGood: I don't think that there's any question that const reference should be used wherever appropriate (but e.g. iterators are usually passed by value, not const reference). My comment was only about *non* const refs.

TallJimbo: @RobertLuptonTheGood could you create an RFC on that?  I'm not opposed, but I don't like deciding coding practices piecemeal by code review.

include/lsst/meas/base/PixelFlags.h
natelust: Removed

include/lsst/meas/base/PixelFlags.h
natelust: updated for clarity

include/lsst/meas/base/PixelFlags.h
natelust: added explicit initialization

include/lsst/meas/base/PixelFlags.h
natelust: removed this functionality

include/lsst/meas/base/PixelFlags.h
natelust: removed the enum

include/lsst/meas/base/PixelFlags.h
natelust: updated syntax

include/lsst/meas/base/PixelFlags.h
natelust: upon discussion, you are right, there is no reasons they needed to be public. They have been moved to private, and the key arrays have been merged together. There is now no different predefined and user specified keymaps. The user specified values are simply added

include/lsst/meas/base/PixelFlags.h
natelust: Updated for references

include/lsst/meas/base/PixelFlags.h
natelust: removed

src/PixelFlags.cc
natelust: changed

src/PixelFlags.cc
natelust: Moved back to anonymous namespace now that the updateflags function was also moved to the anonymous namespace

src/PixelFlags.cc
natelust: updated

src/PixelFlags.cc
natelust: Will be split out to a separate commit, when the code is cleaned up for final merge

src/PixelFlags.cc
natelust: fixed

src/PixelFlags.cc
natelust: updated

src/PixelFlags.cc
natelust: changed to maskName to indicate that it is the name of a mask plane

src/PixelFlags.cc
natelust: fixed and migrated function to anonymous namespace

src/PixelFlags.cc
natelust: updated for better functionality

src/PixelFlags.cc
natelust: wired auto newline in editor, fixed now

src/PixelFlags.cc
natelust: updated

src/PixelFlags.cc
natelust: removed

src/PixelFlags.cc
natelust: This is indeed better syntax and has been updated

src/PixelFlags.cc
natelust: changed to more efficient try catch syntax

src/PixelFlags.cc
natelust: fixed

src/PixelFlags.cc
natelust: fixed

src/PixelFlags.cc
natelust: fixed

src/PixelFlags.cc
natelust: added explanatory comment

src/PixelFlags.cc
natelust: There is no explicit guarantee in the object itself that the first peak is the highest, but all of the measurement framework has been setup in such a way that the list of peaks will be populated with the highest first. Updated the comment to reflect that.

RobertLuptonTheGood: If we're worried we could add a flag that's set when we sort the list.  Or, I suppose, use a heap or some other data structure that guarantees the sort order.

src/PixelFlags.cc
natelust: It is set now

tests/measureSources.py
natelust: I disagree with splitting out these tests into separate tests. Within this file, each test represents testing the functionality of an individual plugin. It would be weird to have each plugin have one test, except pixel flags which has 3. And I don't really see what it gains you. Separate tests are to indicate which portion of code fails (or preferably passes). In this case that would indicate the problem is in pixel flags. To gain more specific information, one can simply look at the specific failure within the test.

include/lsst/meas/base/PixelFlags.h
```diff
@@ -94,10 +79,13 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
     ) const;
 
 private:
-
+    typedef std::map<std::string, afw::table::Key<afw::table::Flag>> KeyMap;
+    typedef afw::image::MaskedImage<float> MaskedImageT;
```
jdswinbank: You don't need `MaskedImageT` here.

tests/measureSources.py
jdswinbank: My suggestion wasn't that they should be separate tests, but that a bit of refactoring within this method could make it shorter and less repetitive (probably by defining a local function and calling that). I agree that it's not worth spending time thinking about though; what you have is fine.

include/lsst/meas/base/PixelFlags.h
```diff
@@ -42,13 +40,19 @@ namespace lsst { namespace meas { namespace base {
...
+                       "List of mask planes to be searched for which occur anywhere within a footprint. "
+                       "If any of the planes are found they will have a corresponding pixel flag set.");
     /**
      *  @brief Default constructor
      *
      *  All control classes should define a default constructor that sets all fields to their default values.
      */
-    PixelFlagsControl() {}
+    PixelFlagsControl():masksFpCenter(), masksFpAnywhere(){
+    }
```
jdswinbank: Please add space around `:` and before `{`, and bring `}` onto the same line.

include/lsst/meas/base/PixelFlags.h
```diff
@@ -94,10 +79,13 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
...
-
+    typedef std::map<std::string, afw::table::Key<afw::table::Flag>> KeyMap;
+    typedef afw::image::MaskedImage<float> MaskedImageT;
     Control _ctrl;
-    FlagHandler _flagHandler;
-    SafeCentroidExtractor _centroidExtractor;
+    KeyMap _centerKeys;
+    KeyMap _anyKeys;
+    afw::table::Key<afw::table::Flag> _generalFailure;
+    afw::table::Key<afw::table::Flag> _offImageKey;
```
jdswinbank: For consistency, shouldn't it be `_generalFailureKey`? (or `_offImage`, with no `Key`, I guess).

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
 private:
     typename MaskedImageT::Mask::Pixel _bits;
 };
-}  // end anonymous namespace
+typedef std::map<std::string, afw::table::Key<afw::table::Flag>> KeyMap;
```
jdswinbank: Please don't `typedef` this twice. You could make it public in the class definition and refer to `PixelFlagsAlgorithm::KeyMap`, perhaps.

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+typedef afw::image::MaskedImage<float> MaskedImageT;
```
jdswinbank: Better not use `MaskedImageT` to refer both to the template parameter above and to the type here. Following coding standard 3-5a, this should be `MaskedImageF`, which has the added benefit of disambiguating them.

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+void updateFlags(KeyMap const & maskFlagToPixelFlag, const FootprintBits<MaskedImageT> & func,
+                                      afw::table::SourceRecord & measRecord){
+        for (auto const & i: maskFlagToPixelFlag){
+            try {
+                if (func.getBits() & MaskedImageT::Mask::getPlaneBitMask(i.first)){
+                    measRecord.set(i.second, true);
+                    }
```
jdswinbank: Watch the indentation.

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+            } 
+            catch (pex::exceptions::InvalidParameterError & err){
+                        throw LSST_EXCEPT(FatalAlgorithmError, err.what());
```
jdswinbank: Too much indentation.

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+    _centerKeys["CR"] = schema.addField<afw::table::Flag>(name + "_flag_crCenter",
+                                        "Cosmic ray in the Source center");
+    _centerKeys["SUSPECT"] = schema.addField<afw::table::Flag>(name + "_flag_suspectCenter",
+                                        "Source's center is close to suspect pixels");
+
+    // Read in the flags passed from the configuration, and add them to the schema
+    for (auto const & i: _ctrl.masksFpCenter){
+        std::string maskName(i);
+        std::transform(maskName.begin(), maskName.end(), maskName.begin(), ::tolower);
+        _centerKeys[i] = schema.addField<afw::table::Flag>(name+"_flag_" + maskName+"Center",
```
jdswinbank: Spacing around the `+` operator.

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+                                        "Source center is close to "+ i + " pixels");
+    }
+    for (auto const & i: _ctrl.masksFpAnywhere){
+        std::string maskName(i);
+        std::transform(maskName.begin(), maskName.end(), maskName.begin(), ::tolower);
+        _anyKeys[i] = schema.addField<afw::table::Flag>(name+"_flag_" + maskName,
```
jdswinbank: Again, be consistent in spacing please.

src/PixelFlags.cc
```diff
@@ -84,60 +130,82 @@ void PixelFlagsAlgorithm::measure(
...
+        //  Catch NAN in centroid estimate
+        if (lsst::utils::isnan(center.getX()) || lsst::utils::isnan(center.getY())) {
+            throw LSST_EXCEPT(pex::exceptions::RuntimeError,
+                              "Center point passed to PixelFlagsAlgorithm is NaN");
+        }
+    }
+    else {
+        // Set the general failure flag because using the Peak might affect
+        // the current measurement
+        measRecord.set(_generalFailure,true);
```
jdswinbank: Space after `,`.

src/PixelFlags.cc
```diff
@@ -84,60 +130,82 @@ void PixelFlagsAlgorithm::measure(
...
-        _flagHandler.setValue(measRecord, SATURATED_CENTER, true);
-    }
-    if (func.getBits() & MaskedImageT::Mask::getPlaneBitMask("CR")) {
-        _flagHandler.setValue(measRecord, CR_CENTER, true);
-    }
-    _flagHandler.setValue(measRecord, FAILURE, false);
+
+    // Update the flags which have to do with the center of the footprint
+    updateFlags(_centerKeys, func, measRecord);
+
```
jdswinbank: Probably don't need this final blank line.

src/PixelFlags.cc
```diff
@@ -84,60 +130,82 @@ void PixelFlagsAlgorithm::measure(
...
 
 void PixelFlagsAlgorithm::fail(afw::table::SourceRecord & measRecord, MeasurementError * error) const {
-    _flagHandler.handleFailure(measRecord, error);
+    measRecord.set(_generalFailure, true);
+    // Because this algorithm does not explicitly throw a MeasurementError we
+    // expect this function to never be executed, and hence error would be
+    // null. If it is run there is a very unexpected behavior and the assert
+    // statement will immediately indicated to the user that something
+    // unexpected has gone wrong 
+    assert (error == NULL);
```
jdswinbank: I'm not sure the above is true, and it's certainly confusing!

First, my understanding is that `fail()` is called for every unhandled exception in the plugin, not just a `MeasurementError`. The fact you don't expect to throw a `MeasurementError` doesn't mean this won't be called.

Second, what does it mean to assert (either in the comment or in the code) that something is `NULL` if it's never going to exist because the function isn't going to be called?

I think you are asserting that any exceptions which are thrown are not `MeasurementError`s. I'm still not really sure why, and the above comment didn't help I'm afraid!

tests/measureSources.py
```diff
@@ -237,15 +237,20 @@ def testPixelFlags(self):
...
         mask.set(20, 20, sat)
         mask.set(60, 60, interp)
         mask.set(40, 20, bad)
         mask.set(20, 80, nodata)
+        mask.set(30, 30, clipped)
         mask.Factory(mask, afwGeom.Box2I(afwGeom.Point2I(0,0), afwGeom.Extent2I(3, height))).set(edge)
         x0, y0 = 1234, 5678
         exp.setXY0(afwGeom.Point2I(x0, y0))
         control = measBase.PixelFlagsControl()
+        # Change the configuration of control to test for clipped mask plane DM-2915
```
jdswinbank: I queried this before, but: is it necessary or helpful to keep repeating the string "DM-2915" in the comments? If every line of code came with a note of the ticket which caused it to be added, it'd get unreadable pretty fast. (Plus, `git blame` would be out of a job.)

src/PixelFlags.cc
```diff
@@ -84,60 +130,82 @@ void PixelFlagsAlgorithm::measure(
...
-    }
-    if (func.getBits() & MaskedImageT::Mask::getPlaneBitMask("SAT")) {
-        _flagHandler.setValue(measRecord, SATURATED, true);
-    }
-    if (func.getBits() & MaskedImageT::Mask::getPlaneBitMask("CR")) {
-        _flagHandler.setValue(measRecord, CR, true);
+
+    // Set the EDGE flag if the bitmask has NO_DATA set
+    try {
+        if (func.getBits() & MaskedImageT::Mask::getPlaneBitMask("NO_DATA")){
```
jdswinbank: Space before `{`.

src/PixelFlags.cc
```diff
@@ -84,60 +130,82 @@ void PixelFlagsAlgorithm::measure(
...
+        // Attempting to set pixel flags with no centroider, the center will
+        // be determined though the peak pixel. The first peak in the
+        // footprint (supplied by the measurement framework) should be the 
+        // highest peak so that one will be used as a proxy for the central 
+        // tendency of the distribution of flux for the record.
+        PTR(afw::detection::Footprint) footprint = measRecord.getFootprint();
+        // If there is no footprint or the footprint contains no peaks, throw
+        // a runtime error.
+        if (!footprint || footprint->getPeaks().empty()){
+            throw LSST_EXCEPT(pex::exceptions::RuntimeError,"No footprint, or no footprint peaks detected");
```
jdswinbank: Space after `,`.

src/PixelFlags.cc
natelust: Jim assures me that for the sake of this function as it is now, we can treat it like it is always sorted by the highest peak, and the other measurements already do. If we want to change the behavior, that is fine, but it is beyond the scope of this port. If you want I can created a ticket for that. Or create a ticket for the general refactoring of the measurement framework and add this to that list.

jdswinbank: There's no requirement to change behaviour as far as I'm concerned; the only requirement is that the behaviour we have should be unambiguous. That's already covered by the changes you've made to the comment.

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+
+typedef afw::image::MaskedImage<float> MaskedImageF;
+
+void updateFlags(PixelFlagsAlgorithm::KeyMap const & maskFlagToPixelFlag, const FootprintBits<MaskedImageF> & func,
+                 afw::table::SourceRecord & measRecord) {
+        for (auto const & i: maskFlagToPixelFlag) {
+            try {
+                if (func.getBits() & MaskedImageF::Mask::getPlaneBitMask(i.first)) {
+                    measRecord.set(i.second, true);
+                    }
```
jdswinbank: Indentation is wrong here (`}` should align with `if`, not `measRecord`).

include/lsst/meas/base/PixelFlags.h
```diff
@@ -42,13 +41,18 @@ namespace lsst { namespace meas { namespace base {
...
+    LSST_CONTROL_FIELD(masksFpAnywhere, std::vector<std::string>,
+                       "List of mask planes to be searched for which occur anywhere within a footprint. "
+                       "If any of the planes are found they will have a corresponding pixel flag set.");
     /**
      *  @brief Default constructor
      *
      *  All control classes should define a default constructor that sets all fields to their default values.
      */
-    PixelFlagsControl() {}
+    PixelFlagsControl() : masksFpCenter(), masksFpAnywhere(){}
```
jdswinbank: Please add a space before `{}`.

src/PixelFlags.cc
```diff
@@ -57,26 +59,70 @@ class FootprintBits : public afw::detection::FootprintFunctor<MaskedImageT> {
...
+            }
+            catch (pex::exceptions::InvalidParameterError & err) {
+                    throw LSST_EXCEPT(FatalAlgorithmError, err.what());
```
jdswinbank: Too much indentation: should be 4 characters, not 8.

PR #27, opened 2015-12-03
Review of DM-4373 (HSC backport: Add tract conveniences)

jdswinbank:


ups/meas_base.table
```diff
@@ -3,6 +3,7 @@ setupRequired(pipe_base)
 setupRequired(daf_base)
 setupRequired(pex_config)
 setupRequired(coadd_utils)
+setupOptional(geom)
```
timj: I'm a bit confused as to how this can be an optional dependency if there is an unprotected import in the code above.

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+import lsst.pex.logging
+import lsst.pex.exceptions
+import lsst.afw.table
+import lsst.afw.image
+import lsst.coadd.utils
+from lsst.geom import convexHull
+
+class PerTractCcdDataIdContainer(lsst.coadd.utils.CoaddDataIdContainer):
+    """A version of lsst.pipe.base.DataIdContainer that combines raw data IDs (defined as whatever we use
+    for 'src') with a tract.
```
jdswinbank: The old `ForcedPhotCcdDataIdContainer` provided a substantially more extensive docstring which explained the motivation. Is it worth preserving (an updated version of) that?

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python
```
jdswinbank: Is this new file really the best location for this code? It rankles a bit since it's cluttering the (already disorganized) `lsst.meas.base` namespace with something that's by no stretch of the imagination a "core measurement algorithm" (as the repository description claims).

As far as I can tell, this is only used by `ForcedPhotCcdTask`. Do you expect it to have wider applicability? If not, can you just fold it into `forcedPhotCcd.py`?

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+import collections
+
+import lsst.pex.logging
+import lsst.pex.exceptions
+import lsst.afw.table
+import lsst.afw.image
+import lsst.coadd.utils
+from lsst.geom import convexHull
+
+class PerTractCcdDataIdContainer(lsst.coadd.utils.CoaddDataIdContainer):
```
jdswinbank: This applies to CCDs, but inherits from `CoaddDataIdContainer`. That seems confusing -- is it necessary?

In fact, I think all you're pulling from `Coadd..` is the `getSkymap()` method. That's really just memoizing `namespace.butler.get(datasetType + "_skyMap")`, and you're also doing that memoization directly (`if skymap is None:`, etc, at line 86 below).

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+    """A version of lsst.pipe.base.DataIdContainer that combines raw data IDs (defined as whatever we use
+    for 'src') with a tract.
+    """
+
+    def castDataIds(self, butler):
```
jdswinbank: This is almost-but-not-quite a copy & paste job on `DataIdContainer.castDataIds()`, which is sad.

At a minimum, it would be nice to describe why that's necessary (the point is that you are augmenting `idKeyTypeDict` to include `tract`/`int`).

It might also be worth considering whether you could reduce the duplication, e.g. by adding an optional `idKeyTypeDict` argument to `DataIdContainer.castDataIds()` and then calling the method in the superclass.

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def makeDataRefList(self, namespace):
+        """Make self.refList from self.idList
+        """
+        if self.datasetType is None:
+            raise RuntimeError("Must call setDatasetType first")
+        skymap = None
+        log = None
+        visitTract = {} # Set of tracts for each visit
+        visitRefs = {} # List of data references for each visit
```
jdswinbank: Need more space before comment on both these lines. (https://confluence.lsstcorp.org/display/LDMDG/Python+Coding+Standard#PythonCodingStandard-Comments)

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+                    log = lsst.pex.logging.Log.getDefaultLog()
+                log.info("Reading WCS for components of dataId=%s to determine tracts" % (dict(dataId),))
+                if skymap is None:
+                    skymap = self.getSkymap(namespace)
+
+                for ref in namespace.butler.subset("calexp", dataId=dataId):
+                    if not ref.datasetExists("calexp"):
+                        continue
+
+                    # XXX fancier mechanism to select an individual exposure than just pulling out "visit"?
```
jdswinbank: If this is (or might be) a requirement, please file a ticket and include the ticket number here. If it's not, please delete the comment.

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+        dataRef = namespace.butler.dataRef(datasetType=self.datasetType, dataId=forcedDataId)
+        self.refList.append(dataRef)
+
+    def makeDataRefList(self, namespace):
+        """Make self.refList from self.idList
+        """
+        if self.datasetType is None:
+            raise RuntimeError("Must call setDatasetType first")
+        skymap = None
+        log = None
```
jdswinbank: Is this dance by which `log` is initialized first to `None` and then only later configured to point to something useful during the core loop really necessary? It makes the code harder to read and sanity check. For example -- is `log` always guaranteed to point to something which has an `info()` method at line 125? (The answer is "yes", but it's not immediately obvious from squinting at the code.)

If setting up logging is an expensive operation, we expect this routine to be called in a tight loop, and that it won't always require logging, this is probably a worthwhile optimization. But I'm pretty sceptical that that's all true.

(A similar-but-milder concern re `skymap`: I can believe that retrieving the skymap from the Butler might be more resource intensive, and the variable isn't referred to again outside the core loop, so it's less dangerous.)

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+                if skymap is None:
+                    skymap = self.getSkymap(namespace)
+
+                for ref in namespace.butler.subset("calexp", dataId=dataId):
+                    if not ref.datasetExists("calexp"):
+                        continue
+
+                    # XXX fancier mechanism to select an individual exposure than just pulling out "visit"?
+                    visit = ref.dataId["visit"]
+                    if visit not in visitRefs:
```
jdswinbank: You could avoid this check by using a `defaultdict`.

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+
+                    md = ref.get("calexp_md", immediate=True)
+                    wcs = lsst.afw.image.makeWcs(md)
+                    box = lsst.afw.geom.Box2D(lsst.afw.geom.Point2D(0, 0),
+                                              lsst.afw.geom.Point2D(md.get("NAXIS1"), md.get("NAXIS2")))
+                    # Going with just the nearest tract.  Since we're throwing all tracts for the visit
+                    # together, this shouldn't be a problem unless the tracts are much smaller than a CCD.
+                    tract = skymap.findTract(wcs.pixelToSky(box.getCenter()))
+                    if overlapsTract(tract, wcs, box):
+                        if visit not in visitTract:
```
jdswinbank: Again, could be a `defaultdict`.

ups/meas_base.table
jdswinbank: Required?

python/lsst/meas/base/dataIds.py
timj: This is something that `flake8` will pick up on where it is consistent with our standard.

python/lsst/meas/base/dataIds.py
```diff
@@ -0,0 +1,153 @@
...
+            raise RuntimeError("Must call setDatasetType first")
+        skymap = None
+        log = None
+        visitTract = {} # Set of tracts for each visit
+        visitRefs = {} # List of data references for each visit
+        for dataId in self.idList:
+            if "tract" not in dataId:
+                # Discover which tracts the data overlaps
+                if log is None:
+                    log = lsst.pex.logging.Log.getDefaultLog()
```
timj: Are we using `pex_logging` here because the rest of the package does and has not yet been ported to `log`?

jdswinbank: More than that: Science Pipelines are explicitly not permitted to use `lsst::log` for now. See [Doxygen](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/log.html) & note that [RFC-29](https://jira.lsstcorp.org/browse/RFC-29) is still just a proposal.

python/lsst/meas/base/dataIds.py
laurenam: It's meant to be used by `meas_mosaic`: see https://github.com/HyperSuprime-Cam/meas_mosaic/search?utf8=✓&q=PerTractCcdDataIdContainer
I tried to keep the import as similar as possible (modulo that the file is now in mea_base, not pipe_tasks).  I think it would be odd to have to call it from `forcedPhotCcd.py`, but if you think there's a more appropriate place for it (`pipe_base` perhaps?), I'm happy to move it.

python/lsst/meas/base/dataIds.py
laurenam: You're absolutely right.  Not necessary at all.  I've adjusted things accordingly in https://github.com/lsst/meas_base/commit/43ffaad8539fd45cf09240d406088d9d440662c9

python/lsst/meas/base/dataIds.py
laurenam: Done.

python/lsst/meas/base/dataIds.py
laurenam: This override turns out to be unnecessary altogether as the `tract/int` pair is being added to the dict through the mapper and policy specifications in `obs_subaru`.  The redundant function has been removed.

python/lsst/meas/base/dataIds.py
laurenam: The dance has been removed in favor of just initializing the log regardless of the presence of a tract argument in the id.

python/lsst/meas/base/dataIds.py
laurenam: Fixed.

python/lsst/meas/base/dataIds.py
laurenam: Line removed.

python/lsst/meas/base/dataIds.py
laurenam: Both defaultdicts added.

ups/meas_base.table
laurenam: Sorry, I blindly followed the HSC entry (which I guess should have been Required).  Amended.

python/lsst/meas/base/dataIds.py
jdswinbank: For what it's worth, I already wrestled with this problem when I started (but still haven't finished...) porting `meas_mosaic` without really finding a solution I was happy with.

If I were you, I would put it in `forcedPhotCcd.py`, since that's the only place it's currently used; if somewhere else would seem more appropriate for `meas_mosaic`, let's worry about that when we do the the `meas_mosaic` port.

That said, I won't strongly object if you decide to leave it in `dataIds.py`. I have no other suggestions as to the location.

laurenam: Moved contents of `dataIds.py` into `forcedPhotCcd.py`.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -35,34 +37,107 @@
...
+                log.info("Reading WCS for components of dataId=%s to determine tracts" % (dict(dataId),))
+                if skymap is None:
+                    skymap = namespace.butler.get(namespace.config.coaddName + "Coadd_skyMap")
+
+                for ref in namespace.butler.subset("calexp", dataId=dataId):
+                    if not ref.datasetExists("calexp"):
+                        continue
+
+                    visit = ref.dataId["visit"]
+                    visitRefs = collections.defaultdict(list)
```
jdswinbank: This looks wrong: every time around the loop you're replacing whatever `visitRefs` used to refer to with a new, empty `defaultdict`. You probably want to move this line outside the loop, replacing line 71.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -35,34 +37,107 @@
...
+
+                    md = ref.get("calexp_md", immediate=True)
+                    wcs = lsst.afw.image.makeWcs(md)
+                    box = lsst.afw.geom.Box2D(lsst.afw.geom.Point2D(0, 0),
+                                              lsst.afw.geom.Point2D(md.get("NAXIS1"), md.get("NAXIS2")))
+                    # Going with just the nearest tract.  Since we're throwing all tracts for the visit
+                    # together, this shouldn't be a problem unless the tracts are much smaller than a CCD.
+                    tract = skymap.findTract(wcs.pixelToSky(box.getCenter()))
+                    if overlapsTract(tract, wcs, box):
+                        visitTract = collections.defaultdict(set)
```
jdswinbank: Same comment as above, but replacing line 70.

laurenam: Again, nice catch.  Both fixed.

PR #28, opened 2015-12-07
Review for DM-4235: countInputs and per object variance functions

jdswinbank:


python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+        maskBits = maskedImage.getMask().getPlaneBitMask(self.config.mask)
+        logicalMask = numpy.logical_not(maskedImage.getMask().getArray() & maskBits)
+        # Compute the median variance value for each pixel not excluded by the mask and write the record
+        medVar = numpy.median(maskedImage.getVariance().getArray()[logicalMask])
+        measRecord.set(self.varValue, medVar)
+
+    def fail(self, measRecord, error=None):
+        measRecord.set(self.varFlag, True)
+
+class SingleFrameCountInputsConfig(SingleFramePluginConfig):
```
jdswinbank: Is `CountInputs` an appropriate name?

I think all the other plugins are named for their output (e.g. `PeakCentroid`, `PsfFlux`, etc) rather than their action (which would give us `CalculatePeakCentroid`, `CountInputs`, etc). Maybe this would be better as `InputCount` or `NumInputs`?

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+    pass
+
+@register("base_CountInputs")
+class SingleFrameCountInputsPlugin(SingleFramePlugin):
+    """
+    Plugin to count how many input images contributed to each source. This information
+    is in the exposure's coaddInputs. Some Limitations:
```
jdswinbank: No need to capitalize "limitations".

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+    * This is only for the pixel containing the center, not for all the pixels in the
+      footprint
```
jdswinbank: But please do capitalize Footprint (and other names of classes).

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+    * This does not account for any clipping in the coadd
+    """
+
+    ConfigClass = SingleFrameCountInputsConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.FLUX_ORDER
```
jdswinbank: Why `FLUX_ORDER`? I think `SHAPE_ORDER` would be more appropriate; according to the documentation, that's for plugins which require `getCentroid()` but not `getShape()`, which is the case here.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.numberKey = schema.addField(name + '_value', type="I",
+                                         doc="Number of images contributing at center; not including any"
```
jdswinbank: Prefer a comma to a semi-colon here.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+                                             "clipping")
+        self.numberFlag = schema.addField(name + '_flag', type="Flag", doc="Set to True for fatal Failure")
+
+    def measure(self, measRecord, exposure):
+        if not exposure.getInfo() or not exposure.getInfo().getCoaddInputs():
+            raise lsst.pex.exceptions.RuntimeError("No coadd inputs defined")
```
jdswinbank: Is it possible to have an `Exposure`  for which `getInfo()` returns `None`? (Genuine question -- I don't think it is, but I may be wrong.)

If `getCoaddInputs()` returns `None`, presumably that means we don't have a coadd. While throwing isn't a terrible idea (the `fail()` method will catch it and set a flag), is there any downside to just recording the number of inputs as `1`?

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+
+        # Promote bounding box of the footprint to tyep D to ensure
+        # the ability to compare (they may be of mixed types I and D)
```
jdswinbank: This comment appears too early -- it refers to line 160, but is immediately followed by line 159 to which it is not relevant.

Also: s/tyep/type/

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+        center = measRecord.getCentroid()
+        fpBbox = lsst.afw.geom.Box2D(measRecord.getFootprint().getBBox())
+        # Check to insure that the center exists and that it is contained within the footprint
```
jdswinbank: s/insure/ensure/

Also, a space between lines 160 and 161 would help readability.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+        # to catch bad centroiding
+        if not center or not fpBbox.contains(center):
+            raise Exception("The measRecord has no center, or the center is outside the footprint")
```
jdswinbank: `measRecord` is just a variable name -- it's an artefact of your implementation, and likely means nothing to the end user. Maybe better to refer to it just as a "measurement" or a "source record" or something?

Also: you can trivially tell if the problem is that the measurement has no centre or if the centre is outside the `Footprint`. The end user likely can't. Would it be worth being explicit about what has caused the failure?

Also also: capital-F-ootprint.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+        measRecord.set(self.varValue, medVar)
+
+    def fail(self, measRecord, error=None):
+        measRecord.set(self.varFlag, True)
+
+class SingleFrameCountInputsConfig(SingleFramePluginConfig):
+    pass
+
+@register("base_CountInputs")
+class SingleFrameCountInputsPlugin(SingleFramePlugin):
```
jdswinbank: Might it also be useful to have a forced version of this plugin?

Actually, I was surprised not to find a Python version of the C++ `SimplePlugin` (ie, a single plugin definition that functions both in single frame & in forced mode). Did I miss it, or do we really not have one?

tests/testCountInputs.py
```diff
@@ -0,0 +1,128 @@
...
+#
+
+"""
+Tests for CountInputs measurement algorithm
+"""
+
+import unittest
+import lsst.utils.tests as utilsTests
+
+import lsst.pex.config as pexConfig
```
jdswinbank: `pexConfig` is unused.

tests/testCountInputs.py
```diff
@@ -0,0 +1,128 @@
...
+import lsst.afw.table as afwTable
+import lsst.afw.image as afwImage
+import lsst.afw.coord as afwCoord
+import lsst.afw.detection as afwDetection
+import lsst.meas.base as measBase
+from lsst.afw.geom.polygon import Polygon
+
+
+class CountInputsTest(unittest.TestCase):
+    def testCountInputs(self):
```
jdswinbank: There's good news and there's bad news here.

The good news: I really like that this test is a serious attempt to exercise the new functionality. Good job!

The bad news: the code is long, confusing and difficult to read.

It doesn't help that the results are a bit unexpected: you have three sources and three input images. One of your sources appears in two inputs, one in one input, and one in none of the inputs. This seems bizarre -- how can you have a source in your coadd that was in none of the inputs?! -- but I'm not sure if it's intentional.

I suggest two specific changes which I think would help:

1. You set `num = 3` at the top of the method, and comment `Number of images`. In fact, it seems like this is _both_ the number of images _and_ the number of test sources. This degeneracy is confusing: if you really want to use `num` for both these quantities, at least say so in the comments. Better still, define two different variables.
2. The code should be better commented throughout. In particular:
  * Add a block at the top, describing the overall aim of the test and summarizing what you expect the results to be. This would be a good place to mention which inputs you expect the sources to appear in, for example;
  * Before each section of the code (setting up the image, setting up the CCDs, adding the sources, running the measurement task), and a brief comment describing what's happening.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+        return cls.FLUX_ORDER
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.varValue = schema.addField(name + '_value', type="D", doc="variance at object position")
+        self.varFlag = schema.addField(name + '_flag', type="Flag", doc="Set to 1 for any fatal Failure")
+
+    def measure(self, measRecord, exposure):
+        # Promote the bounding box of the footprint to type D to ensure
+        # the ability to compare (they may be of mixed types I and D)
```
jdswinbank: This comment appears too early -- it refers to line 103, but is immediately followed by line 102 to which it is not relevant.

Also, I think it would be helpful to specify what you want to compare it with.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+    ConfigClass = SingleFrameVarianceConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.FLUX_ORDER
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.varValue = schema.addField(name + '_value', type="D", doc="variance at object position")
+        self.varFlag = schema.addField(name + '_flag', type="Flag", doc="Set to 1 for any fatal Failure")
```
jdswinbank: To `1` or to `True`? Also, no need to capitalize "failure".

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+
+    ConfigClass = SingleFrameVarianceConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.FLUX_ORDER
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.varValue = schema.addField(name + '_value', type="D", doc="variance at object position")
```
jdswinbank: Capital-V-ariance.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
 
 # --- Single-Frame Measurement Plugins ---
+class SingleFrameVarianceConfig(SingleFramePluginConfig):
+        scale = lsst.pex.config.Field(dtype=float, default=5.0, optional=True,
+                                      doc="scale factor to apply to shape for aperture")
+        mask = lsst.pex.config.ListField(doc="mask planes to ignore", dtype=str,
+                                         default=["DETECTED", "DETECTED_NEGATIVE"])
+
+@register("base_Variance")
+class SingleFrameVariancePlugin(SingleFramePlugin):
```
jdswinbank: Is there any benefit in (or requirement for) a forced measurement plugin?

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+
+    def measure(self, measRecord, exposure):
+        # Promote the bounding box of the footprint to type D to ensure
+        # the ability to compare (they may be of mixed types I and D)
+        center = measRecord.getCentroid()
+        fpBbox = lsst.afw.geom.Box2D(measRecord.getFootprint().getBBox())
+        # check to insure that the center exists and that it is contained within the footprint
+        # catch bad centroiding
+        if not center or not fpBbox.contains(center):
+            raise Exception("The measRecord has no center, or the center is outside the footprint")
```
jdswinbank: measRecord is just a variable name -- it's an artefact of your implementation, and likely means nothing to the end user. Maybe better to refer to it just as a "measurement" or a "source record" or something?

Also: you can trivially tell if the problem is that the measurement has no centre or if the centre is outside the Footprint. The end user likely can't. Would it be worth being explicit about what has caused the failure?

Also also: capital-F-ootprint.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
                     TransformClass=bl.ApertureFluxTransform, executionOrder=BasePlugin.FLUX_ORDER)
 
 # --- Single-Frame Measurement Plugins ---
+class SingleFrameVarianceConfig(SingleFramePluginConfig):
+        scale = lsst.pex.config.Field(dtype=float, default=5.0, optional=True,
+                                      doc="scale factor to apply to shape for aperture")
```
jdswinbank: I personally would start all these `doc=` strings with capitals. However, we aren't consistent about that across the codebase as it is, so I guess it's not a requirement.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
...
+        aperture.scale(self.config.scale)
+        foot = lsst.afw.detection.Footprint(aperture)
+        foot.clipTo(exposure.getBBox(lsst.afw.image.PARENT))
+        # Filter out any pixels which have mask bits set corresponding to the planes to be excluded
+        # (defined in config.mask)
+        maskedImage = exposure.getMaskedImage()
+        maskBits = maskedImage.getMask().getPlaneBitMask(self.config.mask)
+        logicalMask = numpy.logical_not(maskedImage.getMask().getArray() & maskBits)
+        # Compute the median variance value for each pixel not excluded by the mask and write the record
+        medVar = numpy.median(maskedImage.getVariance().getArray()[logicalMask])
```
jdswinbank: Would it be better to use `lsst.afw.math.makeStatistics()` rather than NumPy? We've seen before that the latter is sometimes inconsistent between versions.

python/lsst/meas/base/plugins.py
```diff
@@ -68,6 +71,103 @@
                     TransformClass=bl.ApertureFluxTransform, executionOrder=BasePlugin.FLUX_ORDER)
 
 # --- Single-Frame Measurement Plugins ---
+class SingleFrameVarianceConfig(SingleFramePluginConfig):
+        scale = lsst.pex.config.Field(dtype=float, default=5.0, optional=True,
+                                      doc="scale factor to apply to shape for aperture")
+        mask = lsst.pex.config.ListField(doc="mask planes to ignore", dtype=str,
+                                         default=["DETECTED", "DETECTED_NEGATIVE"])
```
jdswinbank: In the test case, you mask `BAD` and `SAT` instead. Naively, it seems like masking all of the above would be a good idea. Any reason why you don't by default?

tests/testVariance.py
```diff
@@ -0,0 +1,131 @@
...
+    display = False
+
+
+class VarianceTest(unittest.TestCase):
+    def testVariance(self):
+        size = 128 # size of image (pixels)
+        center = afwGeom.Point2D(size//2, size//2) # object center
+        width = 1.0 # PSF width
+        flux = 10.0 # Flux of object
+        variance = 1.0 # Variance value
```
jdswinbank: Setting the variance to be constant across the image limits your capability for testing. You assert below that the value recovered by your plugin is `1.0`, which is fine insofar as it goes, but it's really only showing that you've successfully managed to read the variance from a single pixel, not that you've managed to select all unmasked pixels within your dilated `Footprint` and find the median variance.

It might be a worthwhile improvement to generate a random (or predictable?) variance image and use that to show your plugin really is doing the calculation it's supposed to.

tests/testVariance.py
```diff
@@ -0,0 +1,131 @@
...
+        # Put in some bad pixels to ensure they're ignored
+        for i in range(-5, 6):
+            bad = size//2 + i*width
+            var.getArray()[bad, :] = float("nan")
+            mask.getArray()[bad, :] = mask.getPlaneBitMask("BAD")
+            var.getArray()[:, bad] = float("nan")
+            mask.getArray()[:, bad] = mask.getPlaneBitMask("BAD")
+
+        if display:
+            import lsst.afw.display.ds9 as ds9
```
jdswinbank: This looks like the old-school `afw.display` interface. I think the modern incantation would be more like:
```python
import lsst.afw.display as afwDisplay
afwDisplay.getDisplay().mtv(image, frame=1)
[etc]
```
See also [Doxygen](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/afw_sec_display.html).

python/lsst/meas/base/plugins.py
natelust: changed

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: I'm not sure about this either, I ported the existing behavior from HSC and assumed the check was in there for some reason.

I would say we would prefer to fail rather than record number of inputs as 1 because presumably this plugin would be used assuming the exposure was a coadd. A failure would tell the user that the exposure was not a coadd vs. succeeding with all records being set to 1. This may introduce some unexpected behavior without user being aware.

natelust: After looking at the code I don't think it is possible to return none unless the program is segfaulting somehow, and so the check is probably unnessisary

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: updated

tests/testCountInputs.py
natelust: removed

tests/testCountInputs.py
natelust: I think I ave addressed most of the issues here, but we should discuss further after you have reviewed the changes

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: fixed

python/lsst/meas/base/plugins.py
natelust: I agree with you, so changed

python/lsst/meas/base/plugins.py
natelust: I simply ported the existing behavior, It probably is a good idea to have them all masked. I will make the change unless you can think of some reason not to now.

python/lsst/meas/base/plugins.py
natelust: Im one of two minds here. On one hand I agree with you that it might not fit how the other plugins do things and this might be a good time to "break" the name as it were. On the other hand, I feel it is kind of important to keep the plugin the same as on HSC to cause minimal other breakage. Also count inputs is still very descriptive of what is going on.

jdswinbank: Isn't the API already substantially different as a result of switching to {{meas_base}}? What are we actually preserving?

python/lsst/meas/base/plugins.py
natelust: Although this does change the runtime level as compared with HSC, I think nothing is depending on this to run, so changing levels should be fine, and will making things to be in line with LSST defined run levels.

python/lsst/meas/base/plugins.py
natelust: Oh I guess that is true! I think it would only be important then to make sure people comparing catalogs are aware of the difference

python/lsst/meas/base/plugins.py
natelust: I spoke with Jim about this, and yes we don't have one. I think the thinking was that any serious plugin would probably be written in c++ anyway. That said there are still ways to defining a base class for a particular plugin, and then using multiple inheritance write both single frame and forced plugin. For this plugin at this point I don't think there is a need to do that, and can be part of another ticket if we wish to develop it.

python/lsst/meas/base/plugins.py
natelust: See my note above about forced measurements. It is possible to create them, but that would belong in a different ticket for creating something new vs porting from HSC

python/lsst/meas/base/plugins.py
natelust: I spoke with Paul about this, and he does not see a compelling enough reason to not use numpy's median, especially as it makes the code very readable. However, If you feel very strongly that this needs done I will change it.

RobertLuptonTheGood: In general (and where it matters) it's probably better to use the afwMath version.  For example, if the data is integer the LSST version can do a better job (I don't that it currently does).  If the data's a maskedImage the LSST code can and should use the variance plane as part of the median calculation.

This is neither a vote for nor against changing this piece of code.

jdswinbank: I'd prefer to see `afw.math`, and it's a trivial tweak to the code, but what you have here is adequate: it's not a requirement to change this in order to pass the review.

tests/testCountInputs.py
```diff
@@ -0,0 +1,207 @@
...
+    '''
+    import itertools
+    import numpy as np
+    import matplotlib.pyplot as plt
+    import matplotlib.patches as patches
+
+    ccdVennDiagram(exp)
+
+class CountInputsTest(unittest.TestCase):
+    def testCountInputs(self, showPlot=False):
```
jdswinbank: This is improved and a bit more explicit than it used to be, but it still seems excessively complex for what it does. A few suggestions to simplify it:

* `offset` is immaterial to the test logic; get rid of it.
* Since there are only three sources, it's actually less typing to specify their positions directly, rather than calculating them based on the image size and an offset;
* Since you are specifying the source positions and the CCD positions, you know a priori how many CCDs overlap with each source. I think it's actually a better idea to use this knowledge than to try and perform the calculation by a different route -- it's more explicit, and it guards against making the same mistake in calculating twice.
* Since the algorithm runs entirely on the catalogue data, there's no need to set pixels in the coadd (except for making the illustrative plot, above).

Using the above, I quickly hacked together https://gist.github.com/jdswinbank/bd608ec0a7e0f92db225. I'm not saying you should adopt that wholesale -- indeed, there are several ways it could be improved given a bit more thought -- but I do think it's significantly clearer & simpler than the current code.

tests/testCountInputs.py
```diff
@@ -0,0 +1,207 @@
...
+    over-plotting the given exposure object.
+    exp [in] - (Exposure) The exposure object to plot, must be the product of a coadd
+    Optional:
+    showImage [in] - (Bool) Plot the data in exp in addition to just it's bounding box, default True
+    LegendLocation [in] - (String) Matplotlib legend location code, can be: 'best', 'upper right',
+                             'upper left', 'lower left', 'lower right', 'right', center left',
+                             'center right', 'lower center', 'upper center', 'center'
+    '''
+    # Create the figure object
+    fig = plt.figure()
```
jdswinbank: This fails (`NameError: global name 'plt' is not defined`). Simply importing it inside the calling function isn't adequate.

tests/testCountInputs.py
```diff
@@ -0,0 +1,207 @@
...
+import lsst.pex.config as pexConfig
+import lsst.afw.geom as afwGeom
+import lsst.afw.table as afwTable
+import lsst.afw.image as afwImage
+import lsst.afw.coord as afwCoord
+import lsst.afw.detection as afwDetection
+import lsst.meas.base as measBase
+from lsst.afw.geom.polygon import Polygon
+
+def ccdVennDiagram(exp, showImage=True, legendLocation='best'):
```
jdswinbank: Is this really a Venn diagram? I think it just looks like one. That's a trivial point, though.

Is it possible to create this diagram using the standard `afw.display` code rather than custom matplotlib? That would be more consistent with the rest of the stack. I don't think you should spent significant effort on it, since it's not a fundamental part of the test, though.

tests/testCountInputs.py
```diff
@@ -0,0 +1,207 @@
...
+        plt.colorbar()
+    # Adjust plot parameters and plot
+    plt.gca().relim()
+    plt.gca().autoscale_view()
+    ylim = plt.gca().get_ylim()
+    xlim = plt.gca().get_xlim()
+    plt.gca().set_ylim(1.5*ylim[0], 1.5*ylim[1])
+    plt.gca().set_xlim(1.5*xlim[0], 1.5*xlim[1])
+    plt.legend(loc=legendLocation)
+    fig.canvas.draw()
```
jdswinbank: Don't you need to do something with the figure after you've generated it? `plt.show()` or save it to a file or similar?

tests/testCountInputs.py
```diff
@@ -0,0 +1,207 @@
...
+        
+        task.run(catalog, exp)
+        # Loop over the postions of each star and calculate how many CCD boxes each is in, and compare
+        # this number to what is returned from the SingleFrameMeasurementTask
+        for i, pp in enumerate(positions):
+            number = sum(afwGeom.Box2D(imageBox).contains(wcs.skyToPixel(wcsRef.pixelToSky(pp))) for
+                         wcs in wcsList)
+            self.assertEqual(catalog[i].get("base_CountInputs_value"), number)
+
+        if showPlot:
```
jdswinbank: The convention in the other test cases is to use `if display` together with a try/except block:
```
try:
    display
except NameError:
    display = False
```
Probably a good idea to adopt the same idiom here.

tests/testCountInputs.py
```diff
@@ -0,0 +1,207 @@
...
+    # Use all the built in matplotib line style attributes to create a list of the possible styles
+    linestyles = ['solid','dashed','dashdot','dotted']
+    colors = ['b','g','r','c','m','y','k']
+    # Calculate the cartisian product of the styles, and randomize the order, to help each CCD get
+    # it's own color
+    pcomb = np.random.permutation(list(itertools.product(colors,linestyles)))
+    # Filter out a black solid box, as that will be the style of the given exp object
+    pcomb = pcomb[((pcomb[:,0] == 'k') * (pcomb[:,1] == 'solid')) == False]
+    # Get the image properties
+    origin = lsst.afw.geom.PointD(exp.getXY0())
```
jdswinbank: You've imported `lsst.afw.geom` as `afwGeom`.

python/lsst/meas/base/plugins.py
natelust: Updated

tests/testVariance.py
```diff
@@ -0,0 +1,139 @@
...
+        for i in range(-5, 6):
+            bad = size//2 + i*width
+            var.getArray()[bad, :] = float("nan")
+            mask.getArray()[bad, :] = mask.getPlaneBitMask("BAD")
+            var.getArray()[:, bad] = float("nan")
+            mask.getArray()[:, bad] = mask.getPlaneBitMask("BAD")
+
+        if display:
+            import lsst.afw.display as afwDisplay
+            afwDisplay.getDisplay().mtv(image, frame=1)
```
jdswinbank: Does this work? Don't you need `afwDisplay.getDisplay(1).mtv(image)`?

tests/testCountInputs.py
natelust: addressed

tests/testCountInputs.py
natelust: changed

tests/testCountInputs.py
natelust: I added a show call. I swear I remember Draw used to call show, but it didn't seem to now

tests/testCountInputs.py
natelust: It's not in the strictest sense a Venn diagram, but close enough that ccdVennDiagram conveys the idea of what is going on pretty well.

It probably could be done with afw.display, but at the time I was just writing something quick for my own benefit, and already knew matplotlib well. I only included the code incase it was ever useful for someone else. At this point I don't think it's worth the time to go back an learn afw.display well enough to translate the functionality.

tests/testCountInputs.py
natelust: fixed

tests/testCountInputs.py
natelust: Adopted

tests/testInputCount.py
```diff
@@ -0,0 +1,205 @@
...
+    ylim = plt.gca().get_ylim()
+    xlim = plt.gca().get_xlim()
+    plt.gca().set_ylim(1.5*ylim[0], 1.5*ylim[1])
+    plt.gca().set_xlim(1.5*xlim[0], 1.5*xlim[1])
+    plt.legend(loc=legendLocation)
+    fig.canvas.draw()
+    plt.show()
+
+
+def displayPlot(exp):
```
jdswinbank: This function now seems rather redundant! Any reason not to just call `ccdVennDiagram()` directly?

tests/testInputCount.py
```diff
@@ -0,0 +1,205 @@
...
+    import matplotlib.patches as patches
+except NameError:
+    display = False
+
+
+def ccdVennDiagram(exp, showImage=True, legendLocation='best'):
+    '''
+    Create a figure with the bounding boxes for each of the images which go into a coadd,
+    over-plotting the given exposure object.
+    exp [in] - (Exposure) The exposure object to plot, must be the product of a coadd
```
jdswinbank: Maybe better to use the standard Doxygen `@param` syntax.

PR #29, opened 2015-12-08
Review for DM-4234: Jacobian and focalplane algorithms

jdswinbank:
:warning: This PR is against `tickets/DM-4235` rather than master! :warning:

Note, as of 1/15/16 This is now properly based off master

### DIFF
--- a/python/lsst/meas/base/plugins.py
+++ b/python/lsst/meas/base/plugins.py
@@ -40,6 +40,7 @@ from .wrappers import wrapSimpleAlgorithm
 from .transforms import SimpleCentroidTransform
 
 __all__ = (
+    "SingleFrameFPPositionConfig", "SingleFrameFPPositionPlugin",
     "SingleFrameJacobianConfig", "SingleFrameJacobianPlugin",
     "SingleFrameVarianceConfig", "SingleFrameVariancePlugin",
     "SingleFrameInputCountConfig", "SingleFrameInputCountPlugin",
@@ -73,6 +74,43 @@ wrapSimpleAlgorithm(bl.CircularApertureFluxAlgorithm, needsMetadata=True, Contro
                     TransformClass=bl.ApertureFluxTransform, executionOrder=BasePlugin.FLUX_ORDER)
 
 # --- Single-Frame Measurement Plugins ---
+class SingleFrameFPPositionConfig(SingleFramePluginConfig):
+    pass
+
+@register("base_FPPosition")
+class SingleFrameFPPositionPlugin(SingleFramePlugin):
+    '''
+    Algorithm to calculate the position of a centroid on the focal plane
+    '''
+
+    ConfigClass = SingleFrameFPPositionConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.SHAPE_ORDER
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.focalValue = lsst.afw.table.Point2DKey.addFields(schema, name, "Position on the focal plane",
+                "mm")
+        self.focalFlag = schema.addField(name + "_flag", type="Flag", doc="Set to True for any fatal failure")
+        self.detectorFlag = schema.addField(name + "_missingDetector_flag", type="Flag",
+                                            doc="Set to True if detector object is missing")
+
+    def measure(self, measRecord, exposure):
+        det = exposure.getDetector()
+        if not det:
+            measRecord.set(self.detectorFlag, True)
+            fp = lsst.afw.geom.Point2D(numpy.nan, numpy.nan)
+        else:
+            center = measRecord.getCentroid()
+            posInPix = det.makeCameraPoint(center, lsst.afw.cameraGeom.PIXELS)
+            fp = det.transform(posInPix, lsst.afw.cameraGeom.FOCAL_PLANE).getPoint()
+        measRecord.set(self.focalValue, fp)
+
+    def fail(self, measRecord, error=None):
+        measRecord.set(self.focalFlag, True)
+
 class SingleFrameJacobianConfig(SingleFramePluginConfig):
     pixelScale = lsst.pex.config.Field(dtype=float, default=0.5, doc="Nominal pixel size (arcsec)")
 
--- a/python/lsst/meas/base/plugins.py
+++ b/python/lsst/meas/base/plugins.py
@@ -40,6 +40,7 @@ from .wrappers import wrapSimpleAlgorithm
 from .transforms import SimpleCentroidTransform
 
 __all__ = (
+    "SingleFrameJacobianConfig", "SingleFrameJacobianPlugin",
     "SingleFrameVarianceConfig", "SingleFrameVariancePlugin",
     "SingleFrameInputCountConfig", "SingleFrameInputCountPlugin",
     "SingleFramePeakCentroidConfig", "SingleFramePeakCentroidPlugin", 
@@ -72,6 +73,41 @@ wrapSimpleAlgorithm(bl.CircularApertureFluxAlgorithm, needsMetadata=True, Contro
                     TransformClass=bl.ApertureFluxTransform, executionOrder=BasePlugin.FLUX_ORDER)
 
 # --- Single-Frame Measurement Plugins ---
+class SingleFrameJacobianConfig(SingleFramePluginConfig):
+    pixelScale = lsst.pex.config.Field(dtype=float, default=0.5, doc="Nominal pixel size (arcsec)")
+
+@register("base_Jacobian")
+class SingleFrameJacobianPlugin(SingleFramePlugin):
+    '''
+    Algorithm which computes the Jacobian about a source and computes its ratio with a nominal pixel area.
+    This allows one to compare relative instead of absolute areas of pixels.
+    '''
+
+    ConfigClass = SingleFrameJacobianConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.SHAPE_ORDER
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.jacValue = schema.addField(name + '_value', type="D", doc='Jacobian correction')
+        self.jacFlag = schema.addField(name + '_flag', type='Flag', doc="Set to 1 for any fatal Failure")
+        # Calculate one over the area of a nominal reference pixel
+        self.scale = pow(self.config.pixelScale, -2)
+
+    def measure(self, measRecord, exposure):
+        center = measRecord.getCentroid()
+        # Compute the area of a pixel at the center of a source records centroid, and take the
+        # ratio of that with the defined reference pixel area.
+        result = numpy.abs(self.scale*exposure.getWcs().linearizePixelToSky(center,
+                           lsst.afw.geom.arcseconds).getLinear().computeDeterminant())
+        measRecord.set(self.jacValue, result)
+
+    def fail(self, measRecord, error=None):
+        measRecord.set(self.jacFlag, True)
+
+
 class SingleFrameVarianceConfig(SingleFramePluginConfig):
         scale = lsst.pex.config.Field(dtype=float, default=5.0, optional=True,
                                       doc="Scale factor to apply to shape for aperture")
--- a/tests/testJacobian.py
+++ b/tests/testJacobian.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.    See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+
+import unittest
+
+import lsst.daf.base as dafBase
+import lsst.afw.image as afwImage
+import lsst.utils.tests
+import lsst.meas.base.tests
+
+from lsst.meas.base.tests import AlgorithmTestCase
+from lsst.meas.base.sfm import SingleFrameMeasurementConfig
+
+class JacobianTestCase(AlgorithmTestCase):
+    def setUp(self):
+        # Pick arbitrary numbers to create a detector object, and a synthetic
+        # dataset. The particular numbers have no special meaning to the test
+        # and be anything as long as they are self consistent (i.e. the
+        # fake source is inside the bounding box)
+        self.center = lsst.afw.geom.Point2D(50.1, 49.8)
+        self.bbox = lsst.afw.geom.Box2I(lsst.afw.geom.Point2I(-20, -30),
+                                        lsst.afw.geom.Extent2I(140,160))
+        self.dataset = lsst.meas.base.tests.TestDataset(self.bbox)
+        self.dataset.addSource(100000.0, self.center)
+
+        md = dafBase.PropertyList()
+
+        for k, v in (
+                ("EQUINOX", 2000.0),
+                ("CRPIX1" , 5353.0),
+                ("CRPIX2" , -35.0),
+                ("CD1_1"  , 0.0),
+                ("CD1_2"  , -5.611E-05),
+                ("CD2_1"  , -5.611E-05),
+                ("CD2_2"  , -0.0),
+                ("CRVAL1" , 4.5789875),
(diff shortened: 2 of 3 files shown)


python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
                     TransformClass=bl.ApertureFluxTransform, executionOrder=BasePlugin.FLUX_ORDER)
 
 # --- Single-Frame Measurement Plugins ---
+class SingleFrameFocalPlaneConfig(SingleFramePluginConfig):
+    pass
+
+@register("base_FocalPlane")
+class SingleFrameFocalPlanePlugin(SingleFramePlugin):
```
jdswinbank: Can we change the name, please? `FocalPlanePosition` or `FPPosition`, maybe?

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+    '''
+    Algorithm to calculate position on focal plane
```
jdswinbank: It's worth specifying that this is the position of the centroid.

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+    '''
+
+    ConfigClass = SingleFrameFocalPlaneConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.APCORR_ORDER
```
jdswinbank: Why `APCORR_ORDER`?

It looks like this requires a centroid but not a shape, so I guess `SHAPE_ORDER` would work.

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.focalValue = lsst.afw.table.Point2DKey.addFields(schema, name, "Position on the focal plane",
+                "mm")
+        self.focalFlag = schema.addField(name + "_flag", type="Flag", doc="Set to 1 for any fatal Failure")
```
jdswinbank: No need to capitalize "Failure". And to `1` or to `True`?

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+
+    def measure(self, measRecord, exposure):
+        center = measRecord.getCentroid()
+        det = exposure.getDetector()
+        if not det:
```
jdswinbank: If you don't have a detector, you don't need a centroid. Why not check for the former first, then not bother fetching the centroid if you're not going to use it?

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+            print('in first')
```
jdswinbank: You don't want this `print`.

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+            fp = lsst.afw.geom.Point2D(numpy.nan, numpy.nan)
```
jdswinbank: Maybe set a failure flag here to indicate what happened?

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+        else:
+            posInPix = det.makeCameraPoint(center, lsst.afw.cameraGeom.PIXELS)
+            fp = det.transform(posInPix, lsst.afw.cameraGeom.FOCAL_PLANE).getPoint()
+        measRecord.set(self.focalValue, fp)
+
+    def fail(self, measRecord, error=None):
+        measRecord.set(self.focalFlag, True)
+
+class SingleFrameJacobianConfig(SingleFramePluginConfig):
+    pixelScale = lsst.pex.config.Field(dtype=float, default=0.5, doc="Nominal pixel size (arcsec)")
```
jdswinbank: Is it necessary to have the user define this? It seems like you should be able to use `exposure.getWcs().getPixelScale()` in `SingleFrameJacobianPlugin.measure()` instead.

python/lsst/meas/base/plugins.py
```diff
@@ -71,6 +73,72 @@
...
+class SingleFrameJacobianPlugin(SingleFramePlugin):
+    '''
+    Algorithm to compute the Jacobian
+    '''
+
+    ConfigClass = SingleFrameJacobianConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.APCORR_ORDER
```
jdswinbank: Could be `SHAPE_ORDER`?

python/lsst/meas/base/plugins.py
PaulPrice: The idea is to have a single nominal pixel size for all CCDs.  If you use `exposure.getWcs().getPixelScale()`, then different CCDs can have different nominal pixel sizes, which means the Jacobian will be discontinuous over CCDs; but the whole idea is to make things continuous over the focal plane.

tests/testFocalPlane.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
+#
+# LSST Data Management System
+# Copyright 2008-2013 LSST Corporation.
```
jdswinbank: Update copyright statement.

tests/testFocalPlane.py
```diff
@@ -0,0 +1,74 @@
...
+
+import numpy
+import unittest
+
+import lsst.daf.base as dafBase
+import lsst.afw.image as afwImage
+from lsst.afw.cameraGeom.testUtils import DetectorWrapper
+import lsst.utils.tests
+import lsst.meas.base.tests
+from lsst.meas.base.tests import (AlgorithmTestCase, SingleFramePluginTransformSetupHelper)
```
jdswinbank: `numpy`, `dafBase`, `afwImage` and `SingleFramePluginTransformSetupHelper` are all unused.

Basically aesthetic, but maybe it would look clearer to separate the `import ...` and the `from ...` statements. Also, they should alphabetic.

python/lsst/meas/base/plugins.py
jdswinbank: Thanks for clarifying, @PaulPrice (and kudos to @natelust for suggesting something similar OOB).

In fact, when I read a bit further and thought about what's going on here, I think the real issue is that we need a bit more documentation here. So far as I understand it, this algorithm is not actually storing the Jacobian determinant, but rather the ratio between the determinant as derived from the `Wcs` and the expectation based on the user-supplied pixel size. That's probably makes sense, but it needs to be documented as such.

tests/testFocalPlane.py
```diff
@@ -0,0 +1,74 @@
...
+        self.dataset.addSource(100000.0, self.center)
+        self.dw = DetectorWrapper()
+
+    def tearDown(self):
+        del self.center
+        del self.bbox
+        del self.dataset
+        del self.dw
+
+    def testJacobianPlugin(self):
```
jdswinbank: `test`... what? Please check the name.

tests/testFocalPlane.py
```diff
@@ -0,0 +1,74 @@
...
+import unittest
+
+import lsst.daf.base as dafBase
+import lsst.afw.image as afwImage
+from lsst.afw.cameraGeom.testUtils import DetectorWrapper
+import lsst.utils.tests
+import lsst.meas.base.tests
+from lsst.meas.base.tests import (AlgorithmTestCase, SingleFramePluginTransformSetupHelper)
+
+class FocalPlaneTestCase(AlgorithmTestCase):
```
jdswinbank: This test is basically fine, but it might be even better if you were to spread a few randomly positioned sources over the image and demonstrate you got back the correct focal plane coordinates for each one rather than just this hard-coded special case.

tests/testFocalPlane.py
```diff
@@ -0,0 +1,74 @@
...
+    def setUp(self):
+        self.center = lsst.afw.geom.Point2D(50.1, 49.8)
```
jdswinbank: When hard-coding numbers into the test case, please indicate where they've come from and if they're in some way special. Is `self.center` required to be `Point2D(50.1, 49.8)`? Would `Point2D(49.8, 50.1)` work just as well? What about the bounding box? The source flux? The dataset noise?

tests/testFocalPlane.py
```diff
@@ -0,0 +1,74 @@
...
+        task = self.makeSingleFrameMeasurementTask("base_FocalPlane")
+        exposure, catalog = self.dataset.realize(10.0, task.schema)
+        exposure.setDetector(self.dw.detector)
+        task.run(exposure, catalog)
+        record = catalog[0]
+        self.assertFalse(record.get("base_FocalPlane_flag"))
+        pointKey = lsst.afw.table.Point2DKey(catalog.schema["base_FocalPlane"])
+        point = record.get(pointKey)
+        self.assertAlmostEqual(point.getX(), 1.012, 3)
+        self.assertAlmostEqual(point.getY(), 1.006, 3)
```
jdswinbank: These numbers are correct, but:

* Where they come from isn't transparent -- the casual reader of the code just has to take it on trust that `point.getX()` should be `1.012`.
* They're dependent on the default arguments to the `DetectorWrapper` constructor, which makes the test fragile to changes elsewhere in the codebase.

You could avoid both of these pitfalls by using something like:
```python
self.assertAlmostEqual(point.getX(), self.dw.pixelSize[0] * (0.5 + self.center[0]), 3)
self.assertAlmostEqual(point.getY(), self.dw.pixelSize[1] * (0.5 + self.center[1]), 3)
```

tests/testJacobian.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python
+#
+# LSST Data Management System
+# Copyright 2008-2013 LSST Corporation.
```
jdswinbank: Update copyright.

tests/testJacobian.py
```diff
@@ -0,0 +1,93 @@
...
+#
+
+import numpy
+import unittest
+
+import lsst.daf.base as dafBase
+import lsst.afw.image as afwImage
+import lsst.utils.tests
+import lsst.meas.base.tests
+from lsst.meas.base.tests import (AlgorithmTestCase, SingleFramePluginTransformSetupHelper)
```
jdswinbank: `numpy` and `SingleFramePluginTransformSetupHelper` are unused.

tests/testJacobian.py
```diff
@@ -0,0 +1,93 @@
...
+
+class JacobianTestCase(AlgorithmTestCase):
+    def setUp(self):
```
jdswinbank: Again, there's a lot of "magic numbers" here. How did you choose them? Are they significant?

tests/testJacobian.py
```diff
@@ -0,0 +1,93 @@
...
+        self.center = lsst.afw.geom.Point2D(50.1, 49.8)
+        self.bbox = lsst.afw.geom.Box2I(lsst.afw.geom.Point2I(-20, -30),
+                                        lsst.afw.geom.Extent2I(140,160))
+        self.dataset = lsst.meas.base.tests.TestDataset(self.bbox)
+        self.dataset.addSource(100000.0, self.center)
+
+        md = dafBase.PropertyList()
+
+        for k, v in (
```
jdswinbank: For the purposes of the test, you could omit some of these keys -- you can drop `EQUINOX`, `RADESYS`, or `CDi_j` (or `CDELTn`) without changing the logic. Less code → fewer mistakes.

tests/testJacobian.py
```diff
@@ -0,0 +1,93 @@
...
+        self.wcs = afwImage.makeWcs(md)
+
+    def tearDown(self):
+        del self.center
+        del self.bbox
+        del self.dataset
+        del self.wcs
+
+    def testJacobianPlugin(self):
+        task = self.makeSingleFrameMeasurementTask("base_Jacobian")
```
jdswinbank: `SingleFrameJacobianConfig` defaults to a pixel scale of 0.5 arcsec. You've gone out of your way above to define a `Wcs` based on a pixel scale of ~0.2 arcsec. Why aren't you also setting that in the configuration?

tests/testJacobian.py
```diff
@@ -0,0 +1,93 @@
...
+        exposure, catalog = self.dataset.realize(10.0, task.schema)
+        exposure.setWcs(self.wcs)
+        task.run(exposure, catalog)
+        record = catalog[0]
+        self.assertFalse(record.get("base_Jacobian_flag"))
+        self.assertAlmostEqual(record.get("base_Jacobian_value"), 0.16320294286075868, 4)
```
jdswinbank: Would be approx. 1 if you set the pixel scale as above -- much easier for the casual reader to understand where that number came from and see that it's correct.

python/lsst/meas/base/plugins.py
natelust: changed to FPPosition

python/lsst/meas/base/plugins.py
natelust: I was trying to keep as close to the old run level as possible. I agree with the the new runlevel and is updated

python/lsst/meas/base/plugins.py
natelust: done

python/lsst/meas/base/plugins.py
natelust: done

python/lsst/meas/base/plugins.py
natelust: done

python/lsst/meas/base/plugins.py
natelust: created flag

python/lsst/meas/base/plugins.py
natelust: I have updated the documentation. Hopefully this is a bit clearer

python/lsst/meas/base/plugins.py
natelust: check

tests/testFocalPlane.py
natelust: done

tests/testFocalPlane.py
natelust: cleaned up

tests/testFocalPlane.py
natelust: good catch

tests/testFocalPlane.py
natelust: Added in additional points

tests/testFocalPlane.py
natelust: added in some explanatory comments

tests/testFocalPlane.py
natelust: updated, and adopted

tests/testJacobian.py
natelust: done

tests/testJacobian.py
natelust: removed

tests/testJacobian.py
natelust: Added a note to indicated the particular numeric values are unimportant for this test case

tests/testJacobian.py
natelust: simplified

tests/testJacobian.py
natelust: updated

tests/testJacobian.py
natelust: updated as part of the above comment

python/lsst/meas/base/plugins.py
```diff
@@ -72,6 +74,78 @@
...
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.focalValue = lsst.afw.table.Point2DKey.addFields(schema, name, "Position on the focal plane",
+                "mm")
+        self.focalFlag = schema.addField(name + "_flag", type="Flag", doc="Set to True for any fatal failure")
+        self.detectorFlag = schema.addField(name + "_missingDetector_flag", type="Flag",
+                                            doc="Set to True if detector object is missing")
+
+    def measure(self, measRecord, exposure):
+        det = exposure.getDetector()
+        center = measRecord.getCentroid()
```
jdswinbank: Following my previous comment, you've switched the order of `det =` and `center =`, which is good. But my point was that `if not det` (the line below) is `True`, you never need to fetch the centroid at all -- you could move the call to `getCentroid()` to after the `else`, for example.

tests/testJacobian.py
```diff
@@ -0,0 +1,98 @@
...
+
+from lsst.meas.base.tests import AlgorithmTestCase
+from lsst.meas.base.sfm import SingleFrameMeasurementConfig
+
+class JacobianTestCase(AlgorithmTestCase):
+    def setUp(self):
+        # Pick arbitrary numbers to create a detector object, and a synthetic
+        # dataset. The particular numbers have no special meaning to the test
+        # and be anything as long as they are self consistent (i.e the
+        # fake source is inside the bounding box_
```
jdswinbank: s/_/)./

tests/testJacobian.py
```diff
@@ -0,0 +1,98 @@
...
+    def tearDown(self):
+        del self.center
+        del self.bbox
+        del self.dataset
+        del self.wcs
+
+    def testJacobianPlugin(self):
+        config = SingleFrameMeasurementConfig()
+        config.plugins.names |= ['base_Jacobian']
+        config.plugins['base_Jacobian'].pixelScale = 0.2
```
jdswinbank: For future reference, it'd be nice to add a comment noting that this is chosen to match the `Wcs` you've defined above.

jdswinbank:
Has been merged.

PR #33, opened 2016-04-22
DM-4887: Refactor measurement afterburners into a new plugin system

PaulPrice:


PaulPrice:
I took the liberty of fixing some typos in the commit messages.

python/lsst/meas/base/baseMeasurement.py
```diff
@@ -30,17 +30,17 @@
...
 
 # Exceptions that the measurement tasks should always propagate up to their callers
 FATAL_EXCEPTIONS = (MemoryError, FatalAlgorithmError)
 
-class BasePluginConfig(lsst.pex.config.Config):
-    """!
-    Base class measurement Plugin config classes.
+class BaseMeasurementPluginConfig(BasePluginConfig):
+    """
+    Base config class for all measurement plugins
```
PaulPrice: If you're going to strip the bang (and I don't know if it's required when there are no doxygen markups), then move the next line up to the classic summary line.

python/lsst/meas/base/forcedMeasurement.py
```diff
@@ -51,17 +51,18 @@
 import lsst.pipe.base
 
 from .pluginRegistry import PluginRegistry
-from .baseMeasurement import BasePluginConfig, BasePlugin, BaseMeasurementConfig, BaseMeasurementTask
+from .baseMeasurement import BaseMeasurementPluginConfig, BaseMeasurementPlugin
+from .baseMeasurement import BaseMeasurementConfig, BaseMeasurementTask
```
PaulPrice: Why split this into two lines?  Is it to support a future change?  Or is it just because it's too long?  Suggest either one `import` per line (easy on git to add/subtract) or use parens around the list and just continue the line.

python/lsst/meas/base/pluginsBase.py
```diff
@@ -0,0 +1,100 @@
+import lsst.pex.config
```
PaulPrice: Copyleft?

python/lsst/meas/base/pluginsBase.py
```diff
@@ -0,0 +1,100 @@
...
+    # named class constants for execution order
+    CENTROID_ORDER = 0.0
+    SHAPE_ORDER = 1.0
+    FLUX_ORDER = 2.0
+    APCORR_ORDER = 3.0
+    CLASSIFY_ORDER = 4.0
+    DEFAULT_AFTERBURNER = 5.0
+
+    @classmethod
+    def getExecutionOrder(cls):
```
PaulPrice: Why a method instead of a variable?  Maybe just because that's the way we've always done it?  But I think a variable would be more pythonic (and it can always be methodised with `@property`).

python/lsst/meas/base/pluginsBase.py
```diff
@@ -0,0 +1,100 @@
...
+        If the exception is a MeasurementError, it will be passed as
+        the error argument; in all other cases the error argument will
+        be None, and the failure will be logged by the measurement
+        framework as a warning.
+        """
+        traceback.print_exc()
+        message = ("The algorithm '%s' thinks it cannot fail, but it did; "
+                   "please report this as a bug (the full traceback is above)."
+                   % self.__class__.__name__)
+        raise NotImplementedError(message)
```
PaulPrice: `NotImplementedError` seems like the wrong type.  `RuntimeErrror` might be better.

python/lsst/meas/base/pluginsBase.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @staticmethod
+    def getTransformClass():
```
PaulPrice: Again, I think this would be better as a class variable rather than a method, but maybe the mechanics already in place want a method.

python/lsst/meas/base/sfm.py
```diff
@@ -30,19 +30,20 @@
 """
 
 from .pluginRegistry import PluginRegistry
-from .baseMeasurement import BasePluginConfig, BasePlugin, BaseMeasurementConfig, BaseMeasurementTask
+from .baseMeasurement import BaseMeasurementPluginConfig, BaseMeasurementPlugin
+from .baseMeasurement import BaseMeasurementConfig, BaseMeasurementTask
```
PaulPrice: 4 separate lines or split over a line using parens?

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        self.plugMetadata = plugMetadata
+
+        self.initializePlugins()
+
+    def initializePlugins(self):
+        '''
+        Initialize the plugins according to the configuration.
+        '''
+
+        pluginType = namedtuple('pluginType','single multi')
```
PaulPrice: I think `Struct` is better than `namedtuple`.  There's a bit more setup, but it's a traditional class so you can look at the code and add methods if appropriate.  If you do go with `namedtuple`, note that because it's a class the variable name (and the first argument) should be Capitalised.

    class PluginType(Struct):
        def __init__(self):
            Struct.__init__(self, single=[], multi=[])

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+import lsst.pipe.base
+import lsst.pex.config
+import lsst.daf.base
+
+from .pluginsBase import BasePlugin, BasePluginConfig
+from .pluginRegistry import PluginRegistry
+from .baseLib import FatalAlgorithmError, MeasurementError
+
+# Exceptions that the measurement tasks should always propagate up to their callers
+FATAL_EXCEPTIONS = (MemoryError, FatalAlgorithmError)
```
PaulPrice: What about `SystemExit` and `KeyboardInterrupt`?  I think you only want to catch things that inherit from `Exception` (not `BaseException`).

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @property
+    def PlugType(self):
+        '''
+        This defines if the plugin operates on a single source at a time, or expects the whole catalog.
+        The value defaults to single for a single source, set to multi when the plugin expects the whole
+        catalog. If The plugin is of type multi, the fail method should be implemented to accept the whole
+        catalog. If the plugin is of type the fail method should accept a single source record.
+        '''
+        return 'single'
```
PaulPrice: Use named constants instead of strings?

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        '''
+        raise NotImplementedError()
+
+    @property
+    def PlugType(self):
+        '''
+        This defines if the plugin operates on a single source at a time, or expects the whole catalog.
+        The value defaults to single for a single source, set to multi when the plugin expects the whole
+        catalog. If The plugin is of type multi, the fail method should be implemented to accept the whole
+        catalog. If the plugin is of type the fail method should accept a single source record.
```
PaulPrice: "If the plugin is of type" what?  (I think you're missing a word, "single" probably.)

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        '''
+        return 'single'
+
+    def burn(self, cat, **kwargs):
```
PaulPrice: Cute name, but I think we should have something more boring but descriptive (like "run") so its easily recognisable as the main deal.  I have regretted being cutesy like this in the past.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        Process either a single catalog enter or the whole catalog and produce output defined by the plugin
+
+        @param[in,out] cat  Either a lsst source catalog or a catalog entery depending on the plug type
+                            specified in the classes configuration. Results may be added to new columns,
+                            or existing entries altered.
+        @param[in] kwargs   Any additional kwargs that may be passed through the afterburner task.
+        """
+        raise NotImplementedError()
+
+class AbContext(object):
```
PaulPrice: `AfterburnerContext`

python/lsst/meas/base/pluginsBase.py
PaulPrice: In fact, it seems to me you should either use a custom exception type just for this very context, or you should keep the same type as the original exception (e.g., `type(error)("Some new message")`) so it can be caught by the user by that type.  You should also get the message from the original exception (and I suggest naming the variable `exc` rather than `error` which might be an error code rather than an exception instance) into what you `raise` so all the information is preserved (in python 3 we will use exception chaining).

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        self.cat = cat
+        self.log = log
+
+    def __enter__(self):
+        return
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if exc_type is None:
+            return True
+        if exc_type in FATAL_EXCEPTIONS:
```
PaulPrice: `if issubclass(exc_type, FATAL_EXCEPTIONS):`
and (below)
`elif issubclass(exc_type, MeasurementError):`

As I've stated elsewhere, I think you want to treat anything that inherits from `BaseException` but not from `Exception` as one of the fatal exceptions because those are meant to bring the system down.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        @param[in,out] cat  Either a lsst source catalog or a catalog entery depending on the plug type
+                            specified in the classes configuration. Results may be added to new columns,
+                            or existing entries altered.
+        @param[in] kwargs   Any additional kwargs that may be passed through the afterburner task.
+        """
+        raise NotImplementedError()
+
+class AbContext(object):
+    '''
+    Context manager to handle catching errors that may have been thrown in an afterburner plugin
```
PaulPrice: Add empty line between summary and details.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, schema, plugMetadata=None, **kwargs):
+        """
+        Constructor; only called by derived classes
+
+        @param[in] plugMetaData     An lsst.daf.base.PropertyList that will be filled with metadata
+                                    about the plugins being run. If None, an empty empty PropertyList
+                                    will be created.
+        @param[in] **kwargs         Additional arguments passed to lsst.pipe.base.Task.__init__.
+        """
+        lsst.pipe.base.Task.__init__(self,**kwargs)
```
PaulPrice: Space after comma.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        else:
+            self.log.warn("Error in {}.burn: {}".format(self.plugin.name, exc_value))
+        return True
+
+class AfterburnerConfig(lsst.pex.config.Config):
+    '''
+    Default AfterburnerConfig. Currently this is an empty list, meaning that there are no default plugins run.
+    The config object for each plugin must use this variable to specify the names of all plugins to be run.
+    '''
+    plugins = AfterburnerPlugin.registry.makeField(multi=True, default=[], doc="Plugins to be run and their "
```
PaulPrice: Put `doc` on next line so string can stay together.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+                                                    "configuration")
+    pass
```
PaulPrice: Don't think you need `pass`.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+    def initializePlugins(self):
+        '''
+        Initialize the plugins according to the configuration.
+        '''
+
+        pluginType = namedtuple('pluginType','single multi')
+        self.executionDict = {}
+        # Read the properties for each plugin. Allocate a dictionary entry for each run level. Verify that
+        # the plugins are above the minimum run level for an afterburner plugin. For each run level, the
+        # plugins are sorted into either single record, or multi record groups to later be run appropriately
```
PaulPrice: I think it's worth hoisting this comment into the method docstring.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+            if executionOrder not in self.executionDict:
+                self.executionDict[executionOrder] = pluginType(single=[], multi=[])
+            if PluginClass.getExecutionOrder() >= BasePlugin.DEFAULT_AFTERBURNER:
+                plug = PluginClass(config, name, self.schema, metadata=self.plugMetadata)
+                if plug.PlugType == 'single':
+                    self.executionDict[executionOrder].single.append(plug)
+                elif plug.PlugType == 'multi':
+                    self.executionDict[executionOrder].multi.append(plug)
+            else:
+                raise ValueError("{} has an execution order less than the minimum for an afterburner plugin."
```
PaulPrice: Need a space before the end quote, or the next line will be right up against the period.

python/lsst/meas/base/afterburner.py
```diff
@@ -0,0 +1,183 @@
...
+        '''
+        The entry point for the afterburner task. This method should be called with a reference to a
+        measurement catalog.
+        '''
+        self.callCompute(measCat)
+
+    def callCompute(self, catalog):
+        '''
+        Run each of the plugins on the catalog
+        @param[in] catalog  The catalog on which the plugins will operate
```
PaulPrice: Blank line between summary and parameters.

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
+import numpy as np
+
+import unittest
+import lsst.utils.tests as utilsTests
+
+import lsst.meas.base.afterburner as AB
```
PaulPrice: Don't think the code standards would smile on `AB`.  I've used something like `mbAfterburner` in the past, but dunno what is recommended.  How about just `afterburner`, or just pull out the symbols you need, or import the symbols into `lsst.meas.base` in the module's \_\_init\_\_.py?

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+    '''
+    Afterburner plugin which is guaranteed to fail, testing the failure framework
+    '''
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.DEFAULT_AFTERBURNER
+
+    def __init__(self, config, name, schema, metadata):
+        AB.AfterburnerPlugin.__init__(self, config, name, schema, metadata)
+        self.failKey = schema.addField(name+"_fail", type="Flag", doc="Failure test")
```
PaulPrice: Space either side of operator `+`.

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def burn(self, measRecord):
+        raise MeasurementError("Supposed to fail", 0)
```
PaulPrice: What's the zero for?

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+    writes out the results to the record
+    '''
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.DEFAULT_AFTERBURNER
+
+    def __init__(self, config, name, schema, metadata):
+        AB.AfterburnerPlugin.__init__(self, config, name, schema, metadata)
+        self.failKey = schema.addField(name+"_fail", type="Flag", doc="Failure flag")
+        self.squareKey = schema.addField(name+"_square", type="D", doc="Square of input catalog")
```
PaulPrice: Spaces around `+`.

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+        measRecord.set(self.squareKey, value**2)
+
+    def fail(self, measRecord, error=None):
+        measRecord.set(self.failKey, True)
+
+@register("multiRecordAfterburner")
+class MultiRecordAb(AB.AfterburnerPlugin):
+    '''
+    Afterburner plugin to test the framework in multimode. This plugin takes the whole source catalog at
+    onece, and loops over the catalog internally. The algorithm simply reads a value, cubes it, and writes the
```
PaulPrice: Typo: "onece".

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+    results out to the table
+    '''
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.DEFAULT_AFTERBURNER
+
+    @property
+    def PlugType(self):
```
PaulPrice: Not sure if I've asked before, but why is `PlugType` capitalised instead of `plugType`?

PaulPrice: Why is this a `@property` instead of just a class variable?

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+        return 'multi'
+
+    def __init__(self, config, name, schema, metadata):
+        AB.AfterburnerPlugin.__init__(self, config, name, schema, metadata)
+        self.failKey = schema.addField(name+"_fail", type="Flag", doc="Failure flag")
+        self.cubeKey = schema.addField(name+"_cube", type="D", doc="Cube of input catalog")
```
PaulPrice: Spaces around `+`.

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+        AB.AfterburnerPlugin.__init__(self, config, name, schema, metadata)
+        self.failKey = schema.addField(name+"_fail", type="Flag", doc="Failure flag")
+        self.sqrtKey = schema.addField(name+"_sqrt", type="D", doc="Square root of singleRecord afterburner")
+
+    def burn(self, measRecord):
+        value = measRecord.get("singleRecordAfterburner_square")
+        measRecord.set(self.sqrtKey, value**0.5)
+
+    def fail(self, measRecord, error=None):
+        measRecord.set(self.failKey, True)
```
PaulPrice: Setting a `self.failKey` in the record seems to be a common pattern.  Can you support this in the base class?

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+    from a previous plugin computes a square root, and writes the results to the table.
+    '''
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.DEFAULT_AFTERBURNER + 1
+
+    def __init__(self, config, name, schema, metadata):
+        AB.AfterburnerPlugin.__init__(self, config, name, schema, metadata)
+        self.failKey = schema.addField(name+"_fail", type="Flag", doc="Failure flag")
+        self.sqrtKey = schema.addField(name+"_sqrt", type="D", doc="Square root of singleRecord afterburner")
```
PaulPrice: Spaces around `+`.

tests/testAfterburners.py
TallJimbo: A better way to write the zero might be `FlagHandler.FAILURE`, if that's available in Python (I don't recall whether it is).

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+        startKey = schema.addField("start", type="D")
+        # Instantiate a config object adding each of the above plugins, and use it to create a task
+        abConfig = AB.AfterburnerConfig()
+        abConfig.plugins.names = ["afterburnerFail", "singleRecordAfterburner",
+                "multiRecordAfterburner", "dependentAfterburner"]
+        abTask = AB.AfterburnerTask(schema=schema, config=abConfig)
+        # Create a catalog with five sources as input to the task
+        self.catalog = afwTable.SourceCatalog(schema)
+        self.numObjects = 5
+        for i in range(self.numObjects):
```
PaulPrice: Probably not necessary here, but you should get in the habit of using `catalog.reserve` when you know the number of records you're going to add.

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+            rec = self.catalog.addNew()
+            rec.set("start", float(i+1))
```
PaulPrice: Spaces around `+`.

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+        # Run the afterburner task, outputs will be checked in test methods
+        abTask.run(self.catalog)
+
+    def testFailKeySet(self):
```
PaulPrice: I think you've set this up as a monolithic test, but dividing it up into multiple `test*` functions doesn't match that approach.  It means that the `setUp` is run multiple times, which is unnecessary and inefficient.

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+        abTask = AB.AfterburnerTask(schema=schema, config=abConfig)
+        # Create a catalog with five sources as input to the task
+        self.catalog = afwTable.SourceCatalog(schema)
+        self.numObjects = 5
+        for i in range(self.numObjects):
+            rec = self.catalog.addNew()
+            rec.set("start", float(i+1))
+        # Run the afterburner task, outputs will be checked in test methods
+        abTask.run(self.catalog)
+
```
PaulPrice: I don't see a `tearDown` function, which is usually necessary for the `MemoryTestCase` to pass when you hold various afw classes in the `TestCase` instance.  Maybe that's because nothing in `SourceCatalog` uses `Citizen`?

tests/testAfterburners.py
```diff
@@ -0,0 +1,181 @@
...
+    def testFailKeySet(self):
+        '''
+        Verify the failure flag got set for the plugin expected to fail
+        '''
+        for i in range(self.numObjects):
```
PaulPrice: I suggest:

    self.assertEqual(len(self.catalog), self.numObjects)
    for src in catalog:

python/lsst/meas/base/baseMeasurement.py
natelust: I am not sure why I took that out, added back now

python/lsst/meas/base/pluginsBase.py
natelust: This is used already in many places in the measurement framework already, and I wanted to stay consistent with the base plugin.

python/lsst/meas/base/pluginsBase.py
natelust: done

python/lsst/meas/base/pluginsBase.py
natelust: This exception is not what is being raised if the plugin fails, each plugin gets to decide what to do in that case. This exception is just to inform that the base class method fail was not overloaded in the child plugin class.

python/lsst/meas/base/pluginsBase.py
natelust: Again this is already existing practice, I just moved this class to this file. I should have stated that when I asked for the review.

python/lsst/meas/base/sfm.py
natelust: Done

tests/testAfterburners.py
natelust: updated to afterBuner variable

tests/testAfterburners.py
natelust: done

tests/testAfterburners.py
natelust: Done

tests/testAfterburners.py
natelust: Done

tests/testAfterburners.py
natelust: That is a good point, it was an artifact of testing that I missed changing. It is updated to a class variable called plugType.

tests/testAfterburners.py
natelust: done

tests/testAfterburners.py
natelust: Done

tests/testAfterburners.py
natelust: I used the same in each of these plugins, but different plugins might have different behaviors. I kept the behavior the same as in the base plugins. Each plugin is free to handle these failures as they please, or may not have any general failure flag, but specify some other way of indicating a failure.

tests/testAfterburners.py
natelust: I will remember that going forward

tests/testAfterburners.py
natelust: done

tests/testAfterburners.py
natelust: Everything in this function is properly being garbage collected at this point, so I saw no need to put one in

tests/testAfterburners.py
natelust: updated logic in this and all following tests

PR #34, opened 2016-05-02
Review for DM-5427: Handle empty variance array gracefuly.

jdswinbank:


jdswinbank:
Typos in commit message: "gracefuly", "pixedls".

python/lsst/meas/base/plugins.py
```diff
@@ -198,8 +198,12 @@ def measure(self, measRecord, exposure):
...
         # Numpy median is used here instead of afw.math makeStatistics because of an issue with data types
         # being passed into the C++ layer (DM-2379).
-        medVar = numpy.median(pixels.getVarianceArray()[logicalMask])
-        measRecord.set(self.varValue, medVar)
+        if numpy.sum(logicalMask) > 0:
+            medVar = numpy.median(pixels.getVarianceArray()[logicalMask])
+            measRecord.set(self.varValue, medVar)
+        else:
+            measRecord.set(self.varValue, numpy.NaN)
+            raise bl.MeasurementError("Footprint empty, or all pixels are masked, can't compute median", bl.FlagHandler.FAILURE)
```
jdswinbank: Line too long.

python/lsst/meas/base/plugins.py
```diff
@@ -198,8 +198,12 @@ def measure(self, measRecord, exposure):
         # Compute the median variance value for each pixel not excluded by the mask and write the record.
         # Numpy median is used here instead of afw.math makeStatistics because of an issue with data types
         # being passed into the C++ layer (DM-2379).
-        medVar = numpy.median(pixels.getVarianceArray()[logicalMask])
-        measRecord.set(self.varValue, medVar)
+        if numpy.sum(logicalMask) > 0:
```
jdswinbank: Why not `numpy.any(logicalMask)`? That seems more explicit, and is (presumably) faster than summing for large masks.

python/lsst/meas/base/plugins.py
jdswinbank: Would it be appropriate to define a specific flag for this failure mode?

Is the reference to `bl.FlagHandler.FAILURE` appropriate? In some sense, I guess it is, but I think if we were doing it properly we'd actually have to define `FAILURE_BAD_CENTROID` above as `bl.FlagHandler.FAILURE+1` rather than just assuming it's 1; if we _can_ assume it's 1, then maybe we can just assume this is 0. Maybe a specific reference to `FlagHandler.FAILURE` is helpful to the reader? (On the other hand, I just wasted 5 minutes wondering about it...)

jdswinbank:
In addition to the minor comments on the code itself, please add an appropriate unit test.

tests/testVariance.py
```diff
@@ -127,6 +127,20 @@ def testVariance(self):
...
+
+        # Now set the pixel mask for all pixels to 'BAD' and remeasure.
+        mask.getArray()[:,:] = mask.getPlaneBitMask("BAD")
+        task.run(catalog, exp)
+
+        # The computed variance should be nan and flag_emptyFootprint should have been set since the footprint
+        #has all masked pixels at this point.
+        self.assertTrue(np.isnan(source.get("base_Variance_value")))
+        self.assertTrue(source.get("base_Variance_flag_emptyFootprint"))
+
```
jdswinbank: If it were possible, it would be nice to make this a separate test method -- it gives us more granularity when detecting failures. You could put all the logic which sets up and runs the test in common function and then have two tests which call it.

python/lsst/meas/base/plugins.py
```diff
@@ -198,10 +201,21 @@ def measure(self, measRecord, exposure):
...
+                                      self.FAILURE_EMPTY_FOOTPRINT)
 
     def fail(self, measRecord, error=None):
+        # Check that we have a error object and that it is of type MeasurementError
+        if isinstance(error, bl.MeasurementError):
+            assert error.getFlagBit() in (self.FAILURE_BAD_CENTROID, self.FAILURE_EMPTY_FOOTPRINT)
+            # FAILURE_BAD_CENTROID handled by alias to centroid record.
+            if error.getFlagBit() == self.FAILURE_EMPTY_FOOTPRINT:
+                measRecord.set(self.emptyFootprintFlag, True)
+        measRecord.set(self.varValue, numpy.nan)
```
jdswinbank: Is explicitly setting the value to NaN necessary?

python/lsst/meas/base/plugins.py
```diff
@@ -198,10 +201,21 @@ def measure(self, measRecord, exposure):
         # Compute the median variance value for each pixel not excluded by the mask and write the record.
         # Numpy median is used here instead of afw.math makeStatistics because of an issue with data types
         # being passed into the C++ layer (DM-2379).
-        medVar = numpy.median(pixels.getVarianceArray()[logicalMask])
-        measRecord.set(self.varValue, medVar)
+        if numpy.any(logicalMask):
+            medVar = numpy.median(pixels.getVarianceArray()[logicalMask])
+            measRecord.set(self.varValue, medVar)
+        else:
+            raise bl.MeasurementError("Footprint empty, or all pixels are masked, can't compute median",
```
jdswinbank: Punctuation is a bit funky. I'd drop the first comma altogether and turn the second into a colon.

PR #35, opened 2016-05-04
DM-5323

r-owen:


python/lsst/meas/base/noiseReplacer.py
```diff
@@ -280,7 +280,7 @@ def getNoiseGenerator(self, exposure, noiseImage, noiseMeanVar, exposureId=None)
         if noiseSource == 'meta':
             # check the exposure metadata
             meta = exposure.getMetadata()
-            # this key name correspond to estimateBackground() in detection.py
+            # this key name correspond to SubtractBackgroundTask() in meas_algorithms
             try:
                 bgMean = meta.getAsDouble('BGMEAN')
```
PaulPrice: Ah, someone is actually using the statistics values!  I think they'd be better off doing the calculation themselves, but this needs changing if we axe the statistics calculation in `SubtractBackgroundTask`.

PR #39, opened 2016-05-05
Migrate boost::tuple to std::tuple

pschella:


timj:
Can you split the commit into the `get` changes and the `tuple` changes?

PR #41, opened 2016-05-21
Changes to FlagHandler so that it can be used for Python only plugins

natelust:
Add method to initialize FlagHandler with a vector of FlagDefinitions
    Allow FlagDefinition to be constructed from a pair of strings
    unit test testFlagHandler added

### DIFF
--- a/include/lsst/meas/base/FlagHandler.h
+++ b/include/lsst/meas/base/FlagHandler.h
@@ -43,13 +43,13 @@ struct FlagDefinition {
     FlagDefinition() {
     }
 
-    FlagDefinition(char const * _name, char const * _doc) {
+    FlagDefinition(std::string _name, std::string _doc) {
         name = _name;
         doc = _doc;
     }
 
-    char const * name;
-    char const * doc;
+    std::string name;
+    std::string doc;
 };
 
 /**
--- a/include/lsst/meas/base/FlagHandler.h
+++ b/include/lsst/meas/base/FlagHandler.h
@@ -39,6 +39,15 @@ namespace lsst { namespace meas { namespace base {
     C-strings so we can create these arrays using initializer lists even in C++98.
  */
 struct FlagDefinition {
+
+    FlagDefinition() {
+    }
+
+    FlagDefinition(char const * _name, char const * _doc) {
+        name = _name;
+        doc = _doc;
+    }
+
     char const * name;
     char const * doc;
 };
@@ -117,6 +126,25 @@ public:
         std::string const & prefix,
         FlagDefinition const * begin,
         FlagDefinition const * end
+    );
+
+    /**
+     *  Add Flag fields to a schema, creating a FlagHandler object to manage them.
+     *
+     *  This is the way FlagHandlers will typically be constructed for new algorithms.
+     *
+     *  @param[out]  schema    Schema to which fields should be added.
+     *  @param[in]   prefix    String name of the algorithm or algorithm component.  Field names will
+     *                         be constructed by using schema.join() on this and the flag name from the
+     *                         FlagDefinition array.
+     *  @param[in]   flagDefs  std::vector of FlagDefinitions
+     *
+     *  This variation of addFields is for Python plugins
+     */
+    static FlagHandler addFields(
+        afw::table::Schema & schema,
+        std::string const & prefix,
+        std::vector<FlagDefinition> const * flagDefs
     );
 
     /**
--- a/src/FlagHandler.cc
+++ b/src/FlagHandler.cc
@@ -44,6 +44,24 @@ FlagHandler FlagHandler::addFields(
     return r;
 }
 
+FlagHandler FlagHandler::addFields(
+    afw::table::Schema & schema,
+    std::string const & prefix,
+    std::vector<FlagDefinition> const * flagDefs
+) {
+    FlagHandler r;
+    r._vector.reserve(flagDefs->size());
+    for (unsigned int i = 0; i < flagDefs->size(); i++) {
+        r._vector.push_back(
+            std::make_pair(
+                flagDefs->at(i),
+                schema.addField<afw::table::Flag>(schema.join(prefix, flagDefs->at(i).name), flagDefs->at(i).doc)
+            )
+        );
+    }
+    return r;
+}
+
 FlagHandler::FlagHandler(
     afw::table::SubSchema const & s,
     FlagDefinition const * begin,
@@ -65,5 +83,6 @@ void FlagHandler::handleFailure(afw::table::BaseRecord & record, MeasurementErro
         record.set(_vector[error->getFlagBit()].second, true);
     }
 }
+
 
 }}} // lsst::meas::base
--- a/tests/testFlagHandler.py
+++ b/tests/testFlagHandler.py
@@ -20,7 +20,7 @@
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
-
+import os
 import unittest
 
 import lsst.utils.tests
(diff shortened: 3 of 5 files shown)


include/lsst/meas/base/FlagHandler.h
```diff
@@ -39,6 +39,15 @@ namespace lsst { namespace meas { namespace base {
...
 struct FlagDefinition {
+
+    FlagDefinition() {
+    }
+
+    FlagDefinition(char const * _name, char const * _doc) {
+        name = _name;
+        doc = _doc;
+    }
+
```
natelust: I am slightly worried what is going to happen when you have python strings mapped to c style pointers. If the python strings somehow get cleaned up. What will happen to the pointers. Is it practical to use std::strings, or am I being overly worried?

src/FlagHandler.cc
```diff
@@ -44,6 +44,24 @@ FlagHandler FlagHandler::addFields(
...
+    r._vector.reserve(flagDefs->size());
+    for (unsigned int i = 0; i < flagDefs->size(); i++) {
+        r._vector.push_back(
+            std::make_pair(
+                flagDefs->at(i),
+                schema.addField<afw::table::Flag>(schema.join(prefix, flagDefs->at(i).name), flagDefs->at(i).doc)
+            )
+        );
+    }
+    return r;
```
natelust: Since we are now using c++ 11, please switch this to a range style for loop. It will make the code more readable, and remove all the extra dereferencing ->

src/FlagHandler.cc
```diff
@@ -66,4 +84,5 @@ void FlagHandler::handleFailure(afw::table::BaseRecord & record, MeasurementErro
     }
 }
 
+
```
natelust: Please don't introduce extra diff churn by adding extra spaces if they don't add any clarity

src/FlagHandler.cc
natelust: I actually think this is a weird method name, as it is really an alternative constructor for a flaghandler, only adding schemas as a side effect. However as this mirrors the existing method it probably is appropriate.

tests/testFlagHandler.py
```diff
@@ -0,0 +1,113 @@
...
+
+class FlagHandlerTestCase(AlgorithmTestCase):
+    
+    def testFlagHandler(self):
+        control = lsst.meas.base.GaussianCentroidControl()
+        alg = lsst.meas.base.GaussianCentroidAlgorithm
+        schema = lsst.afw.table.SourceTable.makeMinimalSchema()
+        plugin = alg(control, 'test', schema)
+        cat = lsst.afw.table.SourceCatalog(schema)
+        schema = cat.getSchema()
```
natelust: you already have a reference to the schema object, this line does not seem to be needed

tests/testFlagHandler.py
```diff
@@ -0,0 +1,113 @@
...
+
+        record = catalog.addNew()
+        error = MeasurementError(fh.getDefinition(SECOND).doc, SECOND)
+        fh.handleFailure(record, error.cpp)
+        self.assertTrue(fh.getValue(record, 0))
+        self.assertFalse(fh.getValue(record, 1))
+        self.assertTrue(fh.getValue(record, 2))
+
+        record = catalog.addNew()
+        error = MeasurementError("Custom error message", FIRST)
```
natelust: Why are you setting a custom string, and not testing anything about it?

tests/testFlagHandler.py
```diff
@@ -0,0 +1,113 @@
...
+        for index, flagDef in enumerate(flagDefs):
+           assert(flagDef.name == flagDefs[index].name)
+           assert(flagDef.doc == flagDefs[index].doc)
+
+        catalog = lsst.afw.table.SourceCatalog(schema)
+
+        # Now check to be sure that all of the known failures set the bits correctly
+        record = catalog.addNew()
+        fh.handleFailure(record)
+        self.assertTrue(fh.getValue(record, 0))
```
natelust: Please use the variable FAILURE instead of 0. Likewise for the other variables

tests/testFlagHandler.py
```diff
@@ -0,0 +1,113 @@
...
+        self.assertFalse(fh.getValue(record, 1))
+        self.assertTrue(fh.getValue(record, 2))
+
+        record = catalog.addNew()
+        error = MeasurementError("Custom error message", FIRST)
+        fh.handleFailure(record, error.cpp)
+        self.assertTrue(fh.getValue(record, 0))
+        self.assertTrue(fh.getValue(record, 1))
+        self.assertFalse(fh.getValue(record, 2))
+
```
natelust: It would actually be useful if you wrote a simple plugin that uses the flag handler. This could serve as an example for anyone going to use this in the future. Alternatively you could maybe just create a small example that uses the flag handler.

tests/testFlagHandler.py
```diff
@@ -0,0 +1,113 @@
...
+            FlagDefinition("1st error", "this is the first failure type"),
+            FlagDefinition("2nd error", "this is the second failure type")
+        ]
+        fh = FlagHandler.addFields(schema, "test",
+            FlagDefinitionVector(flagDefs))
+
+        # Check to be sure that the FlagHandler was correctly initialized
+        for index, flagDef in enumerate(flagDefs):
+           assert(flagDef.name == flagDefs[index].name)
+           assert(flagDef.doc == flagDefs[index].doc)
```
natelust: This doesn't actually test what you say it does in the comment, you are actually just comparing an object to itself, as can be seen with id(). I think you are trying to compare to fh.getDefinition[index].name. Please fix the test.

include/lsst/meas/base/FlagHandler.h
pgee2000: Good catch.  I am not sure about the lifetime issue here, but it certainly should be thought about.  Doesn't the same issue come put for the original addFields() call?  Seems like what protects those pointers is that they are attached to a static member of the class which is using them.  That means that someone abusing the original addFields() could probably create the same issue.  But I propose that I adopt the same strategy with Python, and use a static list for the flagDefs and only call the addFields from the plugin constructor.

tests/testFlagHandler.py
pgee2000: Yes, I agree.  I was thinking of the ngmix plugin as the "sample", but it is easy enough to transfer that code to a simple example.

include/lsst/meas/base/FlagHandler.h
TallJimbo: This would indeed cause problems if called from Python.  But the reason we needed to use raw pointers instead of `std::string` (direct initialization of an array of struts containing strings) has gone away in C++.  Please try just switching to `std::string`.

src/FlagHandler.cc
pgee2000: Since this code is a copy of the original addFields, we should probably change both or neither.

tests/testFlagHandler.py
pgee2000: It isn't possible to test this really without an actual plugin, so I removed it.

tests/testFlagHandler.py
pgee2000: I swear that this once was a real test.

src/FlagHandler.cc
pgee2000: I'm leaving it up to Jim whether he wants to make this change.

TallJimbo: Let's not change it.  I think of the side-effect of adding fields as actually being more important than the fact that this constructs a FlagHandler, and the same name is used for similar methods on FunctorKeys in many other places.

include/lsst/meas/base/FlagHandler.h
```diff
@@ -39,8 +39,17 @@ namespace lsst { namespace meas { namespace base {
     C-strings so we can create these arrays using initializer lists even in C++98.
```
jdswinbank: Please update the comment above to match the changes you've made below.

PR #42, opened 2016-06-01
DM-5877: Use Afterburners to clean up aperture correction logic

PaulPrice:


python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+from .afterburner import AfterburnerPluginConfig, AfterburnerPlugin
+from .pluginRegistry import register
+
+__all__ = (
+    "ABClassificationConfig", "ABClassificationPlugin",
+)
+
+
+class ABClassificationConfig(AfterburnerPluginConfig):
+
```
PaulPrice: Remove the blank line since it doesn't add anything and adds potential for the parts being separated.

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+    modelErrFactor = lsst.pex.config.Field(dtype=float, default=0.0, optional=True,
+                                           doc="correction factor for modelFlux error")
+    psfErrFactor = lsst.pex.config.Field(dtype=float, default=0.0, optional=True,
+                                         doc="correction factor for psfFlux error")
+
+
+@register("base_ClassificationExtendedness")
+class ABClassificationPlugin(AfterburnerPlugin):
+    """
+    A binary measure of the extendedness of a source, based a simple cut on the ratio of the
```
PaulPrice: based on

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+    PSF flux to the model flux.
+
+    Because the fluxes on which this algorithm is based are slot measurements, they can be provided
+    by different algorithms, and the "fluxRatio" threshold used by this algorithm should generally
+    be set differently for different algorithms.  To do this, plot the difference between the PSF
+    magnitude and the model magnitude vs. the PSF magnitude, and look for where the cloud of galaxies
+    begins.
```
PaulPrice: Presumably you also have to convert that magnitude difference to a ratio.

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+    fluxRatio = lsst.pex.config.Field(dtype=float, default=.925, optional=True,
+                                      doc="critical ratio of model to psf flux")
+    modelErrFactor = lsst.pex.config.Field(dtype=float, default=0.0, optional=True,
+                                           doc="correction factor for modelFlux error")
+    psfErrFactor = lsst.pex.config.Field(dtype=float, default=0.0, optional=True,
+                                         doc="correction factor for psfFlux error")
+
+
+@register("base_ClassificationExtendedness")
+class ABClassificationPlugin(AfterburnerPlugin):
```
PaulPrice: Drop the leading `AB`.  Standards say it should be spelled out if it's there.

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+
+    ConfigClass = ABClassificationConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.DEFAULT_AFTERBURNER
+
+    def __init__(self, config, name, schema, metadata):
+        AfterburnerPlugin.__init__(self, config, name, schema, metadata)
+        self.keyProbability = schema.addField(name + "_value", type="D",
```
PaulPrice: Please don't call this a `Probability`, since this classifier only does a binary classification.  So then is there any reason (besides history) not to make this a flag?

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+                                              doc="Set to 1 for extended sources, 0 for point sources.")
+        self.keyFlag = schema.addField(name + "_flag", type="Flag", doc="Set to 1 for any fatal failure.")
+
+    def burn(self, measRecord):
+        modelFlux = measRecord.getModelFlux()
+        psfFlux = measRecord.getPsfFlux()
+        modelFluxFlag = (measRecord.getModelFluxFlag()
+                         if measRecord.table.getModelFluxFlagKey().isValid()
+                         else False)
```
PaulPrice: I don't think it's in the coding standards, but I generally advise putting line breaks after an operator (in this case, `if` and `else` are the operators) rather than before because it makes it clearer that there's something missing if a line is accidentally deleted.

python/lsst/meas/base/classification.py
RobertLuptonTheGood: I'm not seeing the total context, but if this is the value (formerly?) known as extendedness it is intended to be a float in the range [0,1].  More a likelihood than a probability, but definitely not a flag.

PaulPrice: That may be the intent, but that is not (nor has it ever been) what it is.  I suggest we either need to change the algorithm to supply a likelihood or change the type to reflect what it is.

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+        psfFluxFlag = (measRecord.getPsfFluxFlag()
+                       if measRecord.table.getPsfFluxFlagKey().isValid()
+                       else False)
+        flux1 = self.config.fluxRatio*modelFlux
+        if not self.config.modelErrFactor == 0:
```
PaulPrice: `x != 0` instead of `not x == 0`.

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+            flux2 += self.config.psfErrFactor*measRecord.getPsfFluxErr()
+
+        # A generic failure occurs when either FluxFlag is set to True
+        # A generic failure also occurs if either calculated flux value is NAN:
+        #     this can occur if the Flux field itself is NAN,
+        #     or the ErrFactor != 0 and the FluxErr is NAN
+        if numpy.isnan(flux1) or numpy.isnan(flux2) or modelFluxFlag or psfFluxFlag:
+            self.fail(measRecord)
+        else:
+            if flux1 < flux2:
```
PaulPrice: This is better written as `measRecord.set(self.keyProbability, 0.0 if flux1 < flux2 else 1.0)`

python/lsst/meas/base/classification.py
```diff
@@ -0,0 +1,104 @@
...
+                measRecord.set(self.keyProbability, 0.0)
+            else:
+                measRecord.set(self.keyProbability, 1.0)
+
+    def fail(self, measRecord, error=None):
+        # Override fail() to do nothing in the case of an exception.  We should be setting a flag
```
PaulPrice: Why a comment instead of a docstring?

tests/testClassification.py
```diff
@@ -64,15 +68,17 @@ def testFlags(self):
...
         config.slots.psfFlux = "base_PsfFlux"
         config.slots.modelFlux = "base_GaussianFlux"
 
+        abConfig = afterburners.AfterburnerConfig()
+
         def runFlagTest(psfFlux=100.0, modelFlux=200.0,
-                       psfFluxSigma=1.0, modelFluxSigma=2.0,
-                       psfFluxFlag=False, modelFluxFlag=False):
+                        psfFluxSigma=1.0, modelFluxSigma=2.0,
+                        psfFluxFlag=False, modelFluxFlag=False):
```
PaulPrice: Make these lines part of the other commit about PEP8 compliance.

tests/testClassification.py
```diff
@@ -121,6 +128,7 @@ def suite():
     suites += unittest.makeSuite(lsst.utils.tests.MemoryTestCase)
     return unittest.TestSuite(suites)
 
+
```
PaulPrice: Remove these extra blank lines.

python/lsst/meas/base/classification.py
RobertLuptonTheGood: Well, for a while in SDSS we did explore how to make this a real float, and José Garmilla has code to convert these psf-model numbers into probabilities now (but we're not using it for LSST).

For now, let's keep it a float and describe it as a probability or likelihood.  @TallJimbo and I are not in favour of making it a bool/flag now as that really is an interpretation of a probability (i.e. involves priors).   If someone wants to start an RFC/c.l.o discussion of this (including adding a slot) that'd be fine, but I think it needs that level of discussion before we start changing things.

python/lsst/meas/base/classification.py
natelust: check

python/lsst/meas/base/classification.py
natelust: check

python/lsst/meas/base/classification.py
natelust: check

python/lsst/meas/base/classification.py
natelust: This was copied from the existing plugin as is, and seems to be correct to my reading

python/lsst/meas/base/classification.py
natelust: keeping as is for now

python/lsst/meas/base/classification.py
natelust: This was moved as is from the previous file and as this is actually a form of comprehension, I think it is reasonable as is.

python/lsst/meas/base/classification.py
natelust: changed

python/lsst/meas/base/classification.py
natelust: changed

python/lsst/meas/base/classification.py
natelust: This comment is actually completely unneeded (this is an override of fail, and plugins themselves are not subclassed to be overridden) I am not sure why it was there in the first place, but has been removed

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -53,11 +55,29 @@ class ProcessImageForcedConfig(lsst.pex.config.Config):
...
+            doc="Subtask to run afterburner plugins on catalog"
+            )
     copyColumns = lsst.pex.config.DictField(
         keytype=str, itemtype=str, doc="Mapping of reference columns to source columns",
         default={"id": "objectId", "parent": "parentObjectId", "deblend_nChild": "deblend_nChild"}
         )
 
+    def setDefaults(self):
+        # Make afterburners a NoOp by default as not modelFlux is setup by default in
+        # ForcedMeasurementTask
```
TallJimbo: "not" - > "no", "NoOp" -> "no-op" (don't want CamelCase to imply it's a class).

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -97,7 +117,11 @@ def __init__(self, butler=None, refSchema=None, **kwds):
         if refSchema is None:
             refSchema = self.references.schema
         self.makeSubtask("measurement", refSchema=refSchema)
-
+        # It is necessary to get the schema internal to the forced measurement task until such a time
+        # that the schema is not owned by the measurement task, but is passed in by an external caller
+        if self.config.doApCorr:
+            self.makeSubtask("applyApCorr", schema=self.measurement.schema)
+        self.makeSubtask('afterburners', schema=self.measurement.schema)
```
TallJimbo: Need a newline after this.

PR #45, opened 2016-07-20
Review for DM-6928: Include PSF moments in the output tables

jdswinbank:


src/SdssShape.cc
```diff
@@ -656,19 +656,33 @@ static std::array<FlagDefinition,SdssShapeAlgorithm::N_FLAGS> const flagDefs = {
...
                                                SIGMA_ONLY);
     r._centroidResult = CentroidResultKey::addFields(schema, name, "elliptical Gaussian adaptive moments",
                                                      NO_UNCERTAINTY);
     r._fluxResult = FluxResultKey::addFields(schema, name, "elliptical Gaussian adaptive moments");
+
+    // Only include PSF shape fields if doMeasurePsf = True
+    if (doMeasurePsf) {
+        r._includePsf = true;
+        r._psfShapeResult = ShapeResultKey::addFields(
+            schema, name + "_psf", "adaptive moments of the PSF model at the object position",
```
TallJimbo: `schema.join(name, "psf")` is slightly preferred over `name + "_psf"`.

src/SdssShape.cc
```diff
@@ -932,13 +992,22 @@ SdssShapeTransform::SdssShapeTransform(
...
         mapper.addMapping(mapper.getInputSchema().find<afw::table::Flag>(
                           mapper.getInputSchema().join(name, flag->name)).key);
     }
 
     _outShapeKey = ShapeResultKey::addFields(mapper.editOutputSchema(), name, "Shape in celestial moments",
                                              FULL_COVARIANCE, afw::table::CoordinateType::CELESTIAL);
+    if (_transformPsf) {
+        _outPsfShapeKey = ShapeResultKey::addFields(mapper.editOutputSchema(), name + "_psf",
+                                                    "PSF shape in celestial moments",
+                                                    FULL_COVARIANCE, afw::table::CoordinateType::CELESTIAL);
```
TallJimbo: I think this should be `NO_UNCERTAINTY` instead of `FULL_COVARIANCE`.

tests/testSdssShape.py
```diff
@@ -127,6 +189,7 @@ def suite():
     suites = []
     suites += unittest.makeSuite(SdssShapeTestCase)
     suites += unittest.makeSuite(SdssShapeTransformTestCase)
+    suites += unittest.makeSuite(PsfSdssShapeTransformTestCase)
```
TallJimbo: This might be a good time to update this test to the [new unit test boilerplate](https://developer.lsst.io/coding/python_testing.html#memory-and-file-descriptor-leak-testing).

TallJimbo:
Did you consider just using `afw::geom::ellipses::Quadrupole` and `afw::table::QuadrupoleKey` instead of `ShapeResult` and `ShapeResultKey`?  I don't see anything obvious that would preclude that, and it would save us from having to put the (unused) uncertainty in the result object.

TallJimbo:
I'm halfway tempted to suggest replacing all of the `(_includePsf ? 0 : 1)` expressions here with just `(!_includePsf)`, since I think those are equivalent.  But I'm only halfway tempted because I'm not sure it's more readable - I'm just bringing it up in case you hadn't considered it but do find it more readable.

jdswinbank:
I like the suggestion of using `Quadrupole` instead of `ShapeResult`. It's an easy change and I don't see any downside. I'll do that.

I find `(!_includePsf)` to be a little obscure: I tend to prefer the current formulation, but I don't feel strongly. I'll leave it as is unless you disagree.

src/SdssShape.cc
jdswinbank: Fixed.

src/SdssShape.cc
jdswinbank: Obsolete due to switch to `QuadrupoleKey`.

tests/testSdssShape.py
jdswinbank: Done. This was [messier than I expected](https://github.com/lsst/meas_base/pull/45/commits/0ee4d5ca15449ae25740ee339408f6d2280b4573), but I think it's worth merging.

PR #47, opened 2016-08-05
DM-6999: Logging framework migration

hchiang2:


isullivan:
It's slightly outside the scope of this pull request, but there are a few lines in these files that could really use to be cleaned up:
in testFlagHandler.py:
* There are a lot of imported packages that are not used
* line 130 there is a "if not X is None:" statement that should be "if X is not None"
* line 131 math is used but not imported

in noiseReplacer.py:
* line 259 there is another "if not X is None:" statement that should be "if X is not None"
* lines 191 and 216 both lines have a "not X in Y" statement that should be "X not in Y"

isullivan:
The code changes all look fine. I don't have the new lsst.log so the changed tests all fail for me, but assuming that they do all pass with the new framework this branch can be merged.

tests/testMeasureSources.py
```diff
@@ -34,6 +34,7 @@
 import numpy as np
 
 import lsst.pex.logging as pexLogging
+from lsst.log import Log
 import lsst.pex.exceptions
```
r-owen: Did you forget to remove this, and the code below that uses it? (I didn't see any other instances, but suggest a global search to be sure).

python/lsst/meas/base/applyApCorr.py
```diff
@@ -194,11 +194,11 @@ def run(self, catalog, apCorrMap):
                 if self.config.doFlagApCorrFailures:
                     source.set(apCorrInfo.fluxFlagKey, oldFluxFlagState)
 
-            if self.log.getThreshold() <= self.log.DEBUG:
+            if self.log.getLevel() <= self.log.TRACE:
```
r-owen: Why the change from debug to trace? In any case this should match the logging level in the conditional (on the copy I just checked out it does match -- I see `self.log.trace` -- but in this pull request it does not -- I see `self.log.debug` -- so that's a bit of a git puzzle).

hchiang2:
Thanks for the pytest work, testFlagHandler.py and noiseReplacer.py have been cleaned up! :-)

python/lsst/meas/base/applyApCorr.py
hchiang2: You are right. I changed it back to DEBUG. 

Also squashed the two commits into one so it's clearer to see this.

tests/testMeasureSources.py
hchiang2: `Log` is used in line 53 below, also added by this commit. 

However, I actually cannot find any logs in `afw` that logs to `afwDetection.Measure`.  It's likely for some old codes that have been replaced?  I will delete these debugging placeholders.

PR #49, opened 2016-08-15
DM-7229: Pytest updates

timj:


tests/testImportCentroidAlgorithm.py
```diff
@@ -59,11 +60,11 @@ def testApplyCentroid(self):
         """Test that we can instantiate and play with SillyMeasureCentroid"""
 
         for imageFactory in (
-                             afwImage.MaskedImageF,
-                             ):
+            afwImage.MaskedImageF,
```
timj: I wonder why this is on a separate line. Were there multiple image types in this test originally?

isullivan: I strongly considered condensing it to one line, but assumed the designer anticipated adding more in the future. I'd be happy to change it.

tests/testMeasureSources.py
```diff
@@ -295,7 +291,7 @@ def testPixelFlags(self):
         source.set("centroid_y", 40)
         source.set("centroid_flag", True)
         source.setFootprint(afwDetection.Footprint(afwGeom.Point2I(afwGeom.Point2D(x + x0, y + y0)), 5))
-        self.assertRaises(  
+        self.assertRaises(
```
timj: @r-owen has suggested we rewrite these using:
```python
with self.assertRaises(lsst.pex.exceptions.RuntimeError):
    plugin.measure(source, exp)
```
I agree with him that this looks much more readable and obvious.

tests/testImportCentroidAlgorithm.py
timj: Leave it be :-)

PR #50, opened 2016-08-23
DM-7262: Port to Python 3

timj:


examples/growthcurve.py
```diff
@@ -100,65 +106,57 @@ def __call__ (self, r):
...
-        kimg   = afwImage.ImageD(kernel.getDimensions())
+        kimg = afwImage.ImageD(kernel.getDimensions())
         kernel.computeImage(kimg, False)
-        kimg  *= 100.0
-        mimg   = afwImage.MaskedImageF(kimg.convertFloat(),
-                                       afwImage.MaskU(kimg.getDimensions(), 0x0),
-                                       afwImage.ImageF(kimg.getDimensions(), 0.0))
+        kimg *= 100.0
+        mimg = afwImage.MaskedImageF(kimg.convertFloat(),
+                                     afwImage.MaskU(kimg.getDimensions(), 0x0),
```
timj: I'm intrigued by the `0x0` here. Doesn't seem to be any different to a straight `0`.

examples/runSingleFrameTask.py
```diff
@@ -22,27 +22,27 @@
...
 from lsst.meas.algorithms.detection import SourceDetectionTask
 from lsst.meas.base import SingleFrameMeasurementTask
 
+
 def loadData():
     """Prepare the data we need to run the example"""
 
     # Load sample input from disk
-    mypath = eups.productDir("afwdata")
+    mypath = getProductDir("afwdata")
```
timj: I'm surprised there was still a `eups.productDir` in there...

python/lsst/meas/base/applyApCorr.py
```diff
@@ -74,32 +77,33 @@ def __init__(self, schema, name):
...
 
+
 class ApplyApCorrConfig(lsst.pex.config.Config):
     ignoreList = lsst.pex.config.ListField(
-        doc = "flux measurement algorithms in getApCorrNameSet() to ignore;" +
-            " if a name is listed that does not appear in getApCorrNameSet() then a warning is logged",
-        dtype = str,
-        optional = False,
-        default = (),
+        doc="flux measurement algorithms in getApCorrNameSet() to ignore;" +
```
timj: I don't think the `+` is needed on this line.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -188,11 +191,11 @@ def makeIdFactory(self, dataRef):
                              (e.g. visit, raft, sensor for LSST data).
         """
         expBits = dataRef.get("ccdExposureId_bits")
-        expId = long(dataRef.get("ccdExposureId"))
+        expId = int(dataRef.get("ccdExposureId"))
```
timj: Is it ok for this to not be `long` on python 2? Noting that on 64-bit python "int" is 64-bit.

python/lsst/meas/base/forcedPhotCoadd.py
```diff
@@ -25,28 +26,29 @@
...
-class ForcedPhotCoaddConfig(ProcessImageForcedConfig):
+
+class ForcedPhotCoaddConfig(ForcedPhotImageConfig):
     footprintDatasetName = lsst.pex.config.Field(
-        doc = ("Dataset (without coadd prefix) that should be used to obtain (Heavy)Footprints for sources."
-               "  Must have IDs that match those of the reference catalog."
-               "  If None, Footprints will be generated by transforming the reference Footprints."),
-        dtype = str,
-        default = "meas",
+        doc=("Dataset (without coadd prefix) that should be used to obtain (Heavy)Footprints for sources."
```
timj: I did a quick test and I don't think the parens are needed here.

python/lsst/meas/base/forcedPhotCoadd.py
```diff
@@ -103,11 +106,11 @@ def makeIdFactory(self, dataRef):
         # allow us to revert back to the old behavior of generating new forced source IDs,
         # just by renaming the ID in config.copyColumns to "object_id".
         expBits = dataRef.get(self.config.coaddName + "CoaddId_bits")
-        expId = long(dataRef.get(self.config.coaddName + "CoaddId"))
+        expId = int(dataRef.get(self.config.coaddName + "CoaddId"))
```
timj: similar long vs int question.

python/lsst/meas/base/pluginsBase.py
```diff
@@ -1,32 +1,35 @@
...
 # GNU General Public License for more details.
-# 
-# You should have received a copy of the LSST License Statement and 
-# the GNU General Public License along with this program.  If not, 
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
+import traceback
```
timj: Is this import needed given that nothing changed in the file to use it.

examples/growthcurve.py
r-owen: Nicely said. I think it was intended to emphasize that this is a mask, whose values are usually set as hex values.

examples/runSingleFrameTask.py
r-owen: Likewise. I thought somebody had eliminated all of them. That reminds me; since utils is directly called in the example I will add it to the required packages (it could be optional, but it must exist anyway so that seems a bit silly)

python/lsst/meas/base/applyApCorr.py
r-owen: Good catch. Also there is a mix of styles for multi-line strings that I standardized.

examples/runSingleFrameTask.py
```diff
@@ -22,27 +22,27 @@
...
     if not mypath:
-        print >> sys.stderr, "Please setup afwdata and try again"
+        print("Please setup afwdata and try again", file=sys.stderr)
         sys.exit(1)
```
timj: This code is incorrect anyhow. It should trigger a test skip if afwdata is missing. It should not simply exit (pytest will break).

timj: Ah. Ok. It's an example not a test file. Ignore me.

python/lsst/meas/base/forcedPhotCcd.py
r-owen: Great catch. I though future had its own version of int to handle this, but I guess not. Ouch!

timj: `from past.builtins import long`

r-owen: It does. The future module ought to have added "from builtins import int" but did not!

timj: Well, then in Python 2 every time you used `int` you would get a `long`. That might also not be desirable so it's not surprising it's not on by default. I think I prefer the `past.builtins` approach even though you end up with code that looks weird on python 3.

python/lsst/meas/base/forcedPhotCoadd.py
r-owen: I agree. I took a cleanup pass on the code to unify multi-line strings (as one commit) and change str % foo -> str % (foo,) as another commit.

python/lsst/meas/base/forcedPhotCcd.py
r-owen: We use 64 bits now, so `int` is safe (all IDs fit in 64 bits) so I propose to leave it as is.

python/lsst/meas/base/pluginsBase.py
r-owen: It was a missing import. traceback is used in the file but was never imported. flake8 exposed the bug.

r-owen: See the fail function defined starting in line 91, in particular line 106:

        traceback.print_exc()

PR #51, opened 2016-09-09
Review for DM-6785: add parent/child measurements

TallJimbo:


python/lsst/meas/base/applyApCorr.py
```diff
@@ -53,38 +53,52 @@ def __init__(self, schema, name):
...
+            self.apCorrKey = schema.addField(
+                name + "_apCorr",
+                doc="aperture correction applied to %s" % (name,),
+                type=float,
+            )
+            self.apCorrSigmaKey = schema.addField(
+                name + "_apCorrSigma",
+                doc="aperture correction applied to %s" % (name,),
+                type=float,
+            )
```
TallJimbo: if `self.doApCorrColumn` is False, we should create an alias that points to the previously-added column (that's exactly the sort of thing aliases were designed for).

python/lsst/meas/base/applyApCorr.py
```diff
@@ -105,6 +126,11 @@ class ApplyApCorrConfig(lsst.pex.config.Config):
         dtype=bool,
         default=True,
     )
+    prefixList = lsst.pex.config.ListField(
+        doc="Prefixes to search for to find additional fields to be aperture-corrected",
+        dtype=str,
+        default=["undeblended_",]
+    )
```
TallJimbo: I'm not sure I like conflating "having a prefix" with "uses another flux to define its aperture corrections".

Could we make this a `DictField` that maps "to-be-corrected" flux to "flux-to-use-for-correction"?  Or better yet, have algorithms specify the flux that should be used to correct them when they're registered?

Side note: this discussion gives me a slight feeling of deja vu, like we already discussed this.  If we did, and we already agreed on the approach you have here, just tell me (though I'd appreciate a reminder of how we got here, or a link to that discussion).  It's also quite possible I just had a conversation about a similar configuration option with someone else.

python/lsst/meas/base/applyApCorr.py
```diff
@@ -148,12 +180,13 @@ def run(self, catalog, apCorrMap):
             self.log.info("Use complex flux sigma computation that double-counts photon noise "
                           "and thus over-estimates flux uncertainty")
```
TallJimbo: While we're here...could you change the above two log lines to `debug` level?  This seems a bit chatty.

tests/testMeasureSources.py
```diff
@@ -300,6 +300,125 @@ def testPixelFlags(self):
...
+        checkSource(child2, fieldName, flux2*apCorrValue)
+
+        # Undeblended
+        checkSource(child1, "undeblended_" + fieldName, (flux1 + flux2)*apCorrValue)
+        checkSource(child2, "undeblended_" + fieldName, (flux1 + flux2)*apCorrValue)
+
+        self.assertIn(fieldName + "_apCorr", schema)
+        self.assertIn(fieldName + "_apCorrSigma", schema)
+        self.assertIn("undeblended_" + fieldName + "_apCorr", schema)
+        self.assertIn("undeblended_" + fieldName + "_apCorrSigma", schema)
```
TallJimbo: Would you mind moving this to a new file and a new TestCase?  `testMeasureSources.py` is a bit of a smorgasbord, and while it's out of scope to fix that here I'd rather not make it worse.

PR #52, opened 2016-09-09
DM-6815: Update LSST full-stack processing configuration to match best practice from HSC

PaulPrice:


python/lsst/meas/base/noiseReplacer.py
```diff
@@ -382,7 +382,7 @@ class ImageNoiseGenerator(NoiseGenerator):
 
     def __init__(self, img):
         """
-        img: an afwImage.ImageF
```
PaulPrice: These docstrings should have a `!` if you're going to use doxygen.  Given the changes coming down the track, I don't feel the need to doxygenate here.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -143,7 +143,7 @@ class ForcedPhotCcdConfig(ForcedPhotImageConfig):
     doApplyUberCal = lsst.pex.config.Field(
         dtype=bool,
         doc="Apply meas_mosaic ubercal results to input calexps?",
-        default=False
+        default=True
```
PaulPrice: Is this going to work for any cameras other than HSC where we have used meas_mosaic?

python/lsst/meas/base/noiseReplacer.py
laurenam: I realize changes are coming, but was just trying to make the file internally consistent (but forgot the !).  I can revert if you think that's preferable.

python/lsst/meas/base/forcedPhotCcd.py
laurenam: I am not sure if this will eventually also apply to the `jointcal` calibration, but no, this currently won't work if there are no `meas_mosaic` results, which have only been run on HSC data (I'm not sure how difficult it would be to get it working on other cameras, nor if there is any real desire to put in that effort given that `jointcal` is intended to be the adopted `ubercal` code).

Note that this was your recommendation (and no one opposed)...are you having second thoughts?  This would mean any non `meas_mosaic`'d processing would need to add **doApplyUberCal=False** to their command lines (be it **makeCoaddTempExp.py** or **coaddDriver.py**) or, alternatively, as config overrides in their **config/makeCoaddTempExp.py** (which we could, in principle, do as part of this ticket).  Let me know if you lean in a particular direction.

python/lsst/meas/base/noiseReplacer.py
PaulPrice: I'm happy to go with whatever you like.

python/lsst/meas/base/forcedPhotCcd.py
PaulPrice: Perhaps I wasn't in my right mind or I didn't understand the import of what I was recommending.

Let's make `doApplyUberCal=False` the default for the stack as a whole, and activate it in obs_subaru only.  We can revisit this choice once @parejkoj and friends are done with jointcal.

parejkoj: 1. doApplyUberCal=False should definitely be the default for the whole stack.
1. I don't think we should call it ubercal, if it's not ubercal (which it isn't).
1. meas_mosiac only works on HSC. jointcal works on cfht and hsc (and sort of on lsstSim), and we're currently working on making it work on general camera data (metadata handling is the problem).

laurenam: Ok, making `doApplyUberCal=False` the default and will override to True in `obs_subaru`.

TallJimbo: I agree with all of the above (including John's suggestion to rename this option, but I think that bit is out of scope for this issue).

PR #57, opened 2016-11-10
meas_base/DM-8221: Save the correct schema in forced photometry

TallJimbo:
Trivial change; note that diff is relative to DM-8210, which was necessary to get the test code (in pipe_tasks) working.  I will rebase after DM-8210 is merged.


TallJimbo:
Closing after manual merge; PR base is not master.

PR #58, opened 2016-11-14
meas_base/DM-8230: updates to forced photometry

TallJimbo:
Two not-obviously-related changes here (on separate commits):
 - We remove a workaround in forced photometry data ID mangling that is now fixed in obs_base
 - We add a config option to support skipping missing patches in forced photometry, which we'll use to enable CCD forced photometry in `ci_hsc` without processing more coadd patches there.

### DIFF
--- a/python/lsst/meas/base/forcedPhotCcd.py
+++ b/python/lsst/meas/base/forcedPhotCcd.py
@@ -54,13 +54,6 @@ class PerTractCcdDataIdContainer(lsst.pipe.base.DataIdContainer):
     the calexp are used.
     """
 
-    def _addDataRef(self, namespace, dataId, tract):
-        """Construct a dataRef based on dataId, but with an added tract key"""
-        forcedDataId = dataId.copy()
-        forcedDataId['tract'] = tract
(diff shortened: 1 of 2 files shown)


TallJimbo:
Note that this diff is relative to DM-8221, which also includes some forced photometry updates.  I'll rebase after review is complete.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -94,16 +87,14 @@ def makeDataRefList(self, namespace):
                     if overlapsTract(tract, wcs, box):
                         visitTract[visit].add(tract.getId())
             else:
-                tract = dataId.pop("tract")
                 # making a DataRef for src fills out any missing keys and allows us to iterate
-                for ref in namespace.butler.subset("src", dataId=dataId):
-                    self._addDataRef(namespace, ref.dataId, tract)
+                self.refList.extend(ref for ref in namespace.butler.subset(self.datasetType, dataId=dataId))
```
PaulPrice: You're not doing this on `"src"` now.  If that's deliberate, remove the comment.

TallJimbo:
Merged manually; PR base was not master.

PR #62, opened 2017-01-27
DM-8461: Wrap pipe_tasks with pybind11

kfindeisen:
Without this import, the docstring created for `SdssCentroidTransform` uses C++ names for the parameter types rather than python names.

### DIFF
--- a/python/lsst/meas/base/baseLib.py
+++ b/python/lsst/meas/base/baseLib.py
@@ -1,5 +1,8 @@
 from __future__ import absolute_import
 
+# Needed for pybind11-generated docstrings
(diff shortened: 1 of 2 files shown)


kfindeisen:
No; as stated in the PR description, the import was needed only to get the right docstring (e.g., `lsst.afw.image._wcs.Wcs` instead of `lsst::afw::image::Wcs`). Would something like "Let pybind11 use python types in documentation" get the message across better?

r-owen:
How about putting a comment to that effect before the import. The commit message is less important, but you might as well update it to match.

kfindeisen:
Updated.

PR #66, opened 2017-02-16
DM-9392: Wrap meas_extensions_convolved with pybind11 instead of swig

pschella:


pschella:
I will perform both changes as part of the cleanup. Otherwise there are too many conflicts.

PR #67, opened 2017-02-24
DM-8465: wrap meas_modelfit with pybind11

TallJimbo:
Add wrappers for some methods apparently not used until meas_modelfit.

### DIFF
--- a/python/lsst/meas/base/sdssShape.cc
+++ b/python/lsst/meas/base/sdssShape.cc
@@ -40,6 +40,32 @@ namespace afwGeom = lsst::afw::geom;
 namespace lsst {
 namespace meas {
 namespace base {
+namespace {
+
+using PySdssShapeAlgorithm = py::class_<SdssShapeAlgorithm, SimpleAlgorithm>;
+
+template <typename ImageT>
+static void declareComputeMethods(PySdssShapeAlgorithm & cls) {
+    cls.def_static(
+        "computeAdaptiveMoments",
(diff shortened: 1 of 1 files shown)


python/lsst/meas/base/sdssShape.cc
```diff
@@ -40,6 +40,32 @@ namespace afwGeom = lsst::afw::geom;
 namespace lsst {
 namespace meas {
 namespace base {
+namespace {
+
+using PySdssShapeAlgorithm = py::class_<SdssShapeAlgorithm, SimpleAlgorithm>;
```
pschella: No `shared_ptr` holder type on purpose?

python/lsst/meas/base/sdssShape.cc
```diff
@@ -40,6 +40,32 @@ namespace afwGeom = lsst::afw::geom;
...
+
+template <typename ImageT>
+void declareComputeMethods(PySdssShapeAlgorithm & cls) {
```
pschella: `static`? Or did we decide that unnamed namespace is enough after all?

python/lsst/meas/base/sdssShape.cc
TallJimbo: Unintentional.  But I'm guessing if I add it here and not in `Algorithm` and all of its derived classes, I'll get segfaults, right?  If so I'd rather defer this to the cleanup issue.

PR #72, opened 2017-03-02
DM-9188: Cleanup pybind11 code in core meas and ip packages

pschella:


python/lsst/meas/base/plugins.py
```diff
@@ -31,13 +31,48 @@
...
+    ScaledApertureFluxControl,
+    ScaledApertureFluxTransform,
+    SdssCentroidAlgorithm,
+    SdssCentroidControl,
+    SdssCentroidTransform,
+    SdssShapeAlgorithm,
+    SdssShapeControl,
+    SdssShapeTransform,
+    )
+
```
TallJimbo: As unpleasant as it is, these probably also ought to be imported from their subpackages/modules instead of from `.`.

PR #74, opened 2017-03-04
Tickets/dm 9711: alternate pull request that is against my rebased DM-8467 = u/rowen/DM-8467

r-owen:


include/lsst/meas/base/CentroidUtilities.h
```diff
@@ -44,21 +44,21 @@ struct CentroidResult {
     CentroidResult();
 
     /// Constructor; initializes everything from values.
-    explicit CentroidResult(CentroidElement _x, CentroidElement _y, CentroidCov const & matrix):
-        x(_x),
-        y(_y),
+    explicit CentroidResult(CentroidElement x, CentroidElement y, CentroidCov const & matrix):
+        x(x),
+        y(y),
```
TallJimbo: I don't think we want to change this, or if we do, we want to change the arguments to e.g. `x_` rather than `x`.  The Python arguments can still be `x`, `y`, of course, but I think it's best to avoid any confusing about which `x` and `y` are in play here.

python/lsst/meas/base/apertureFlux.cc
```diff
@@ -19,6 +19,7 @@
  * the GNU General Public License along with this program.  If not, 
  * see <https://www.lsstcorp.org/LegalNotices/>.
  */
+#include <memory>
```
TallJimbo: pybind11 headers should go first.

python/lsst/meas/base/blendedness.cc
```diff
@@ -19,6 +19,7 @@
  * the GNU General Public License along with this program.  If not, 
  * see <https://www.lsstcorp.org/LegalNotices/>.
  */
+#include <memory>
```
TallJimbo: pybind11 headers first.

python/lsst/meas/base/flagHandler.cc
```diff
@@ -32,36 +34,90 @@ namespace lsst {
...
+
+    cls.def(py::init<>());
+    cls.def(py::init<std::initializer_list<FlagDefinition> const &>());
+
+    cls.def("__getitem__",
+            [](FlagDefinitionList const &self, int i) {
+                auto cind = lsst::utils::python::cppIndex(self.size(), i);
+                return self[cind];
+            },
+            py::is_operator());
```
TallJimbo: No reason to use `py::is_operator` for anything but binary operators.

include/lsst/meas/base/CentroidUtilities.h
r-owen: I strongly disagree (clearly, else I would not have made the change). I personally hate arguments with needless underscores and in a situation like this where the arguments are being used to attributes that have the same name, using the same name seems intuitive and natural to me. I realize it can be a bit confusing to a newbie, but since the results are obvious I see that as harmless.

python/lsst/meas/base/flagHandler.cc
r-owen: Are you sure? I thought it would result in `__getitem__` giving the expected behavior if the type of the index-like argument was wrong (e.g. a string provided when an int was expected).

TallJimbo: Hmm. I haven't tried it, but I assumed it just makes it return `NotImplemented` when there's an exception thrown, and I think for `__getitem__` that would just be returned to the user.

TallJimbo: Confirmed; user just gets `NotImplemented` returned.

TallJimbo:
I believe this has long been merged to master via another branch/PR.

PR #76, opened 2017-03-20
DM-9190: Cleanup pybind11 remaining code

pschella:


python/lsst/meas/base/sincCoeffs.cc
```diff
@@ -52,4 +52,6 @@ PYBIND11_PLUGIN(sincCoeffs) {
     return mod.ptr();
```
r-owen: I believe this module needs to import `lsst.afw.geom` and `lsst.afw.image`

PR #77, opened 2017-03-31
DM-10039: Incorrect docs for CatalogCalculationConfig.

jdswinbank:


python/lsst/meas/base/catalogCalculation.py
```diff
@@ -103,9 +103,9 @@ def __exit__(self, exc_type, exc_value, traceback):
 
 class CatalogCalculationConfig(lsst.pex.config.Config):
     '''
-    Default CatalogCalculationConfig. Currently this is an empty list, meaning that there are no default
```
natelust: This doc string did not get updated when a default plugin was added (on 5877 I think) Im sorry I didn't catch it, but am happy you did

python/lsst/meas/base/catalogCalculation.py
```diff
@@ -103,9 +103,9 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-    plugins run. The config object for each plugin must use this variable to specify the names of all
-    plugins to be run.
+    Config class for catalog calculation driver task.
+
+    Specifies which plugins our catalog calculation will use.
```
natelust: I would prefer the text: " Specifies which plugins will execute when a CatalogCalculationTask is run"

PR #79, opened 2017-04-13
Tickets/dm 8108

natelust:


src/PsfFlux.cc
```diff
@@ -106,25 +111,14 @@ void PsfFluxAlgorithm::measure(
     typedef afw::detection::Psf::Pixel PsfPixel;
     typedef afw::image::MaskedImage<float>::Variance::Pixel VarPixel;
     ndarray::EigenView<PsfPixel,1,1,Eigen::ArrayXpr> model(
-        afw::detection::flattenArray(
-            fitRegion,
-            psfImage->getArray(),
-            psfImage->getXY0()
-        )
+        fitRegion.getSpans()->flatten(psfImage->getArray(), psfImage->getXY0())
```
parejkoj: That's an improvement.

src/PsfFlux.cc
```diff
@@ -94,7 +94,12 @@ void PsfFluxAlgorithm::measure(
         ) {
             badBits |= exposure.getMaskedImage().getMask()->getPlaneBitMask(*i);
         }
-        fitRegion.intersectMask(*exposure.getMaskedImage().getMask(), badBits);
+        auto tmpSpans = afw::geom::SpanSet::fromMask(*exposure.getMaskedImage().getMask(),
+                                                     [&badBits]
+                                                     (afw::image::MaskPixel const & bitPattern)
+                                                     {return bitPattern & badBits;});
+        fitRegion.setSpans(fitRegion.getSpans()->intersectNot(*tmpSpans)->clippedTo(
+            exposure.getMaskedImage().getMask()->getBBox()));
```
parejkoj: Another case where the syntax seems much more clumsy than before.

tests/testMeasureSources.py
```diff
@@ -286,7 +290,9 @@ def testPixelFlags(self):
         with self.assertRaises(lsst.pex.exceptions.RuntimeError):
             plugin.measure(source, exp)
         # The second test will raise an error because no peaks are present
-        source.setFootprint(afwDetection.Footprint(afwGeom.Point2I(afwGeom.Point2D(x + x0, y + y0)), 5))
+        tmpSpanSet2 = afwGeom.SpanSet.fromShape(5).shiftedBy(x + x0,
+                                                             y + y0)
+        source.setFootprint(afwDetection.Footprint(tmpSpanSet2))
```
parejkoj: Ah the joys of 79 characters...

tests/test_undeblended.py
```diff
@@ -107,14 +111,13 @@ def testUndeblendedMeasurement(self):
...
+        spanSetChild2 = spanSetChild2.shiftedBy(*tmpPoint)
+        foot2 = afwDetection.Footprint(spanSetChild2)
         child2.setFootprint(afwDetection.HeavyFootprintF(foot2, image))
 
-        spans = []
-        for ss in foot1.getSpans():
-            spans.append(ss)
-        for ss in foot2.getSpans():
-            spans.append(ss)
+        spans = foot1.spans.union(foot2.spans)
```
parejkoj: That's an improvement.

src/PixelFlags.cc
```diff
@@ -174,7 +169,7 @@ void PixelFlagsAlgorithm::measure(
 
     // Check for bits set in the source's Footprint
     afw::detection::Footprint const & footprint(*measRecord.getFootprint());
-    func.apply(footprint);
+    footprint.getSpans()->clippedTo(mimage.getBBox())->applyFunctor(func, *(mimage.getMask()));
```
natelust: This new syntax exists for two reasons. 1. Most of this sort of work is attempted to be part of the new SpanSet class, such that in many places people can just use SpanSets, and not carry around the extra bits of footprints. I attempted to change the minimal amount in the stack to ensure maximum compatibility, and thus I did not go all in on where SpanSets could be used exclusively. Future code would probably not be used in this manor. 2. Previously the footprint functors silently dropped locations that were in the footprint but outside images, arrays, etc. This is a dangerous thing, but was commonly exploited by code that dilated the footprint such that the bounds were outside the image. For better safety it is now required that the programmer explicitly acknowledge that they wanted these pixels dropped with a clippedTo call.

parejkoj: Ok, that's fair.

src/PsfFlux.cc
natelust: I'm glad you pointed this out, as it was partly left over from an intermediate development stage. The code now reads:
`fitRegion.setSpans(fitRegion.getSpans()->intersectNot(*exposure.getMaskedImage().getMask(),
            badBits)->clippedTo(exposure.getMaskedImage().getMask()->getBBox()))`
If you ignore the fact that this is sitting inside a footprint, this is exactly the same syntax but now on a spanset object. This SpanSet just must be accessed, and set to, the footprint object, as it itself is immutable. We may consider an api update in the future to add a convenience method to do with from the footprint if it proves necessary. We are hoping that SpanSets will take over in newly developed code such that this sort of thing will be quite rare.

parejkoj: If you think an API update is worth it, please file a ticket now so we don't forget. If you suspect it will likely be unneeded, don't bother with the ticket.

PR #80, opened 2017-04-20
DM-10237: fix errors in blendedness calculation

TallJimbo:
This addresses three problems in the blendedness calculation:
 - a simple transcription bug in what is now computeAbsExpectation (this is what originally spawened the ticket);
 - the integral in what is now computeAbsExpectation was over a nonnormalized probability;
 - we were truncating de-biased absolute values before summing them, which was introducing another bias we weren't correcting for.

I've also added some new unit tests.

### DIFF
--- a/src/Blendedness.cc
+++ b/src/Blendedness.cc
@@ -210,7 +210,7 @@ void computeMoments(
             float variance = pixelIter.variance();
             float mu = BlendednessAlgorithm::computeAbsExpectation(data, variance);
             float bias = BlendednessAlgorithm::computeAbsBias(mu, variance);
-            accumulatorAbs(d.getX(), d.getY(), weight, std::max(std::abs(data) - bias, 0.0f));
+            accumulatorAbs(d.getX(), d.getY(), weight, std::abs(data) - bias);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Blendedness.cc
```diff
@@ -284,6 +279,22 @@ BlendednessAlgorithm::BlendednessAlgorithm(Control const & ctrl,  std::string co
     }
 }
 
+float BlendednessAlgorithm::computeAbsExpectation(float data, float variance) {
+    float normalization = 0.5f*boost::math::erfc(-data/std::sqrt(2.0f*variance));
```
yalsayyad: std::erfc available with C++11?

TallJimbo: It hadn't occurred to me to check, but it looks like it is.  I'll update all the code using the boost version in this file.

tests/testBlendedness.py
```diff
@@ -24,6 +24,8 @@
 from contextlib import contextmanager
 import unittest
 
+import numpy as np
```
yalsayyad: Leftover import from debugging?

TallJimbo: From an older form of the test, yes.

PR #82, opened 2017-04-20
DM-10287 Add and enable a plugin to record Footprint area.

TallJimbo:


python/lsst/meas/base/plugins.py
```diff
@@ -371,6 +372,38 @@ def fail(self, measRecord, error=None):
...
+        self.key = schema.addField(
+            schema.join(name, "value"),
+            type=numpy.int32,
+            doc="Number of pixels in the source's detection footprint.",
+            units="pixel"
+        )
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.CENTROID_ORDER  # probably doesn't need to go early, but can
```
natelust: I don't think we are supposed to do comments after statements. Additionally the wording of the comment is a bit awkward, and required re-reading to notice it was not cut short somehow

TallJimbo: If comments are statements are not allowed it's news to me (but it may have been a change brought on by adopting more of PEP8, which I haven't followed closely). In any case, this is now moot since I've rewritten this as a `CatalogCalculation` and I'll just use the default execution order for those.

PR #89, opened 2017-08-08
DM-11514: Fix pytest execution

timj:
* Ignore non-test python files.
* Correct name for setup_module.
* Also rename test files.


tests/SConscript
```diff
@@ -2,6 +2,9 @@
 from lsst.sconsUtils import scripts
 
 pybind11_test_modules = ['sillyCentroid',]
+ignoreList = [name + '.py' for name in pybind11_test_modules]
+ignoreList.append('testLib.py')
 
 scripts.BasicSConscript.pybind11(pybind11_test_modules)
-scripts.BasicSConscript.tests(noBuildList=[name + '.cc' for name in pybind11_test_modules])
+scripts.BasicSConscript.tests(noBuildList=[name + '.cc' for name in pybind11_test_modules],
+                              ignoreList=ignoreList)
```
kfindeisen: Perhaps make the code more symmetric, so it's easier to read?

timj:
I'll annotate jira. The problem is the work on the ticket has not yet finished (sconsUtils) but the work like this can be merged early.

tests/SConscript
timj: you mean move the `noBuildList` list building outside? Ok.

PR #90, opened 2017-08-12
DM-11539: Fix test for pytest and make flake8 clean

timj:
Ensures that the dummy classes registered in one test are still valid so that a later test will not complain about them.

Also make flake8 clean and fix a couple of warnings. The remaining flake8 warnings are caused by Doxygen `##` syntax. @jonathansick I am assuming that they will be fixed as part of the migration to numpydoc.

### DIFF
--- a/python/lsst/meas/base/forcedPhotCoadd.py
+++ b/python/lsst/meas/base/forcedPhotCoadd.py
@@ -1,4 +1,3 @@
-from builtins import zip
 #!/usr/bin/env python
 #
 # LSST Data Management System
@@ -21,6 +20,8 @@ from builtins import zip
 # the GNU General Public License along with this program.  If not,
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
+from builtins import zip
+
 import lsst.pex.config
 import lsst.pipe.base
 import lsst.coadd.utils
--- a/tests/test_ApertureFlux.py
+++ b/tests/test_ApertureFlux.py
@@ -70,7 +70,7 @@ class ApertureFluxTestCase(lsst.utils.tests.TestCase):
                                                          position)
                 area = self.computeNaiveArea(position, radius)
                 # test that this isn't the same as the sinc flux
-                self.assertNotClose(
+                self.assertFloatsNotEqual(
                     ApertureFluxAlgorithm.computeSincFlux(self.exposure.getMaskedImage().getImage(),
                                                           ellipse, self.ctrl).flux, area)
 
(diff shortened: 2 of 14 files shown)


jonathansick:
That's right.

tests/test_PluginLogs.py
```diff
@@ -208,7 +208,7 @@ def testLoggingPythonPlugin(self):
         pluginLogName = os.path.join(lsst.utils.getPackageDir('meas_base'), 'tests', 'plugin.log')
         directLog(log, pluginLogName)
         exposure, cat = self.dataset.realize(noise=0.0, schema=schema)
-        source = cat[0]
+        # source = cat[0]  # noqa F841: Not sure if this is deliberate
```
jdswinbank: Can we become sure, rather than just adding the note? If it doesn't break when you take it out, it looks very much like an oversight to me.

tests/test_PluginLogs.py
```diff
@@ -235,7 +235,7 @@ def testLoggingCppPlugin(self):
...
```
jdswinbank: Ditto.

python/lsst/meas/base/forcedPhotCoadd.py
```diff
@@ -52,8 +53,8 @@ def setDefaults(self):
 
     def validate(self):
         ForcedPhotImageTask.ConfigClass.validate(self)
-        if (self.measurement.doReplaceWithNoise and self.footprintDatasetName is not None
-                and self.references.removePatchOverlaps):
+        if (self.measurement.doReplaceWithNoise and self.footprintDatasetName is not None and
+                self.references.removePatchOverlaps):
```
jdswinbank: I'm confused by this. I see that you're squelching a `flake8` warning, but you actually seem to be going against the [PEP8 recommendation](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) (and against Knuth, which I personally think is much more significant). Is this a known issue in `flake8`? Has it been discussed elsewhere?

tests/test_ApertureFlux.py
```diff
@@ -70,7 +70,7 @@ def testNaive(self):
                                                          position)
                 area = self.computeNaiveArea(position, radius)
                 # test that this isn't the same as the sinc flux
-                self.assertNotClose(
+                self.assertFloatsNotEqual(
```
jdswinbank: Wasn't immediately clear why you did this — I had to go and look up the definition of `assertNotClose` to find out that it's deprecated. Maybe worth noting that in the commit message?

tests/test_PluginLogs.py
timj: I have removed them as the test passes regardless. My comment was really for the reviewer.

python/lsst/meas/base/forcedPhotCoadd.py
timj: It was W503. Weirdly, `pycodestyle` implies that W503 is disabled by default (this was changed in the past year). I am using the current version so it's almost like my `flake8` is forcing it to be enabled. I've added it to the ignore list explicitly. I wasn't able to install the newest flake8 as an eups package because of some problem with `setuptools` version numbering.

PR #92, opened 2017-08-19
Use getTempFilePath to use unique log names

timj:


timj:
This was a minor change directly related to DM-11514 that I was going to sort out so I left the work on the ticket. It wouldn't be needed if we weren't trying to get DM-11514 merged. The reason DM-11514 was "Reviewed" was that it was reviewed before people asked for `xdist` support to be added. That multi-process change led to all these additional problems. Thanks for the review.

PR #94, opened 2017-08-28
DM-11620: Numerical problem with Gaussian test in meas_base with pytest

laurenam:


tests/test_SdssCentroid.py
```diff
@@ -96,7 +95,7 @@ def testMonteCarlo(self):
             ySigmaList = []
             nSamples = 1000
             for repeat in range(nSamples):
-                exposure, catalog = self.dataset.realize(noise*flux, schema)
+                exposure, catalog = self.dataset.realize(noise*flux, schema, randomSeed=repeat)
```
timj: Add a comment as for the other two times?

PR #95, opened 2017-08-29
DM-10902: Stack demo failure with devtoolset-6

pschella:


timj:
We use `float` surprisingly often in `meas_base`.

pschella:
Yes, and we probably shouldn't. Just because it mostly operates on `Image<float>` that doesn't mean it is a good idea to use `float` for all intermediate values. However, tracking where we should and shouldn't is beyond the scope of this ticket I think.

timj:
I'm happy with the fix here. I wasn't suggesting we audit `meas_base` now (maybe we should file a ticket though).

PR #99, opened 2017-10-17
DM-12243: make inputCount more robust

PaulPrice:
And run it in forced coadd measurement.

### DIFF
--- a/python/lsst/meas/base/forcedPhotCoadd.py
+++ b/python/lsst/meas/base/forcedPhotCoadd.py
@@ -50,7 +50,7 @@ class ForcedPhotCoaddConfig(ForcedPhotImageConfig):
         self.measurement.copyColumns["id"] = "id"
         self.measurement.copyColumns["parent"] = "parent"
         self.references.removePatchOverlaps = False  # see validate() for why
-        self.measurement.plugins.names |= ['base_InputCount']
+        self.measurement.plugins.names |= ['base_InputCount', 'base_Variance']
 
     def validate(self):
         ForcedPhotImageTask.ConfigClass.validate(self)
--- a/python/lsst/meas/base/forcedPhotCoadd.py
+++ b/python/lsst/meas/base/forcedPhotCoadd.py
@@ -50,6 +50,7 @@ class ForcedPhotCoaddConfig(ForcedPhotImageConfig):
         self.measurement.copyColumns["id"] = "id"
         self.measurement.copyColumns["parent"] = "parent"
         self.references.removePatchOverlaps = False  # see validate() for why
+        self.measurement.plugins.names |= ['base_InputCount']
 
     def validate(self):
         ForcedPhotImageTask.ConfigClass.validate(self)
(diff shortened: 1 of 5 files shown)


python/lsst/meas/base/wrappers.py
```diff
@@ -442,3 +444,201 @@ def _init(self, ctrl, name, mapper, logName=None):
...
+            self._generic.fail(measRecord, error if error is not None else None)
+
+        @classmethod
+        def getExecutionOrder(cls):
+            return GenericPluginClass.getExecutionOrder()
+
+        def getTransformClass(self):
+            return self._generic.getTransformClass()
+
+    return ForcedFromGenericPlugin
```
natelust: I think I would prefer to see these as class methods on the generic plugin class called something like make<Single/forced>Plugin, so that something that is wrapped can just be singleFooPlugin = genericFooPluging.makeSingle("name"), each of these could then even deffer to a third more common code base to reduce duplication further if desired

python/lsst/meas/base/wrappers.py
```diff
@@ -442,3 +444,201 @@ def _init(self, ctrl, name, mapper, logName=None):
...
+    class SingleFrameFromGenericConfig(GenericPluginClass.ConfigClass, SingleFramePluginConfig):
+        pass
+
+    @register(name)
+    class SingleFrameFromGenericPlugin(SingleFramePlugin):
+        ConfigClass = SingleFrameFromGenericConfig
+
+        def __init__(self, config, name, schema, metadata, logName=None):
+            SingleFramePlugin.__init__(self, config, name, schema, metadata, logName=logName)
+            self._generic = GenericPluginClass(config, name, schema, metadata)
```
natelust: I am nearly positive that if you are creating these classes inside a function the GenericPluginClass (or cls; see below) variable will be in a closure, and there is no need to explicitly store another pointer to it inside the self object.

python/lsst/meas/base/forcedPhotCoadd.py
```diff
@@ -50,6 +50,7 @@ def setDefaults(self):
         self.measurement.copyColumns["id"] = "id"
         self.measurement.copyColumns["parent"] = "parent"
         self.references.removePatchOverlaps = False  # see validate() for why
+        self.measurement.plugins.names |= ['base_InputCount', 'base_Variance']
```
natelust: why do these get promoted to always be run, as opposed to camera configs or run specific configs? Is this something we thing everyone will want all the time?

PaulPrice: Certainly the `InputCount` is something everyone will want to use (e.g., it can be used as a poor-man's detection limit). HSC experience is that the `Variance` is helpful too.

python/lsst/meas/base/wrappers.py
PaulPrice: I'm storing an instance, not the class itself. I don't think we want to instantiate it every time we want to use it.
Or have I misunderstood your point?

python/lsst/meas/base/wrappers.py
PaulPrice: Sounds like a good idea. I'll give it a go.

PaulPrice:
Moved the factory functions into the `GenericPlugin` as `classmethod`s, and rebased against master.
Rerunning Jenkins...

PR #105, opened 2017-12-19
DM-13055 Add centroid CovarianceMatrix to TestDataset

parejkoj:


python/lsst/meas/base/tests.py
```diff
@@ -345,6 +348,9 @@ def addSource(self, flux, centroid, shape=None):
         record = self.catalog.addNew()
         record.set(self.keys["flux"], flux)
         record.set(self.keys["centroid"], centroid)
+        offdiag = centroid[0]*centroid[1]*0.001
+        covariance = np.matrix(((centroid[0]*0.01, offdiag), (offdiag, centroid[1]*0.01)))
+        record.set(self.keys["centroid_sigma"], covariance.astype(np.float32))
```
TallJimbo: Setting the centroid uncertainty to be proportional to the centroid is a little weird, as it gives the uncertainty a trend with position.  Of course, the test data is highly artificial, and if you want such a trend to test something, I think that's fine - just wanted to see what you were thinking here.

parejkoj: No deep thought process, I just needed to put *something* in for the errors. If you have something else you'd prefer, please let me know.

TallJimbo: If there's already an RNG near this code, maybe just make them random?  If not, don't worry about it.

parejkoj: Take a look at the change I just pushed and tell me what you think.

PR #107, opened 2018-02-13
DM-10957: Add measurement algorithm for point sources on fluff

PaulPrice:


include/lsst/meas/base/LocalBackground.h
```diff
@@ -0,0 +1,107 @@
+// -*- lsst-c++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2008-2017 AURA/LSST.
```
ctslater: Might as well say 2018 in the boilerplates, fwiw.

PR #108, opened 2018-02-22
DM-13615: TestDataset allows randomness

jdswinbank:


python/lsst/meas/base/tests.py
```diff
@@ -197,9 +197,6 @@ def makePerturbedWcs(oldWcs, minScaleFactor=1.2, maxScaleFactor=1.5,
         The default range for rotation is 30-60 degrees, and the default range for reference shift
         is 0.5-1.0 arcseconds (these cannot be safely included directly as default values because Angle
         objects are mutable).
-
-        The random number generator is primed with the seed given. If ``None``, a seed is
-        automatically chosen.
```
jdswinbank: This is still true, isn't it? I understand that you don't want `randomSeed=None` in general use, and I agree with defaulting it to something else, but that doesn't invalidate the statement you're removing.

PR #118, opened 2018-06-07
DM-14725: Eliminate explicit use of ndarray::EigenView in C++ code

r-owen:


src/PsfFlux.cc
```diff
@@ -93,13 +93,15 @@ void PsfFluxAlgorithm::measure(afw::table::SourceRecord& measRecord,
...
-    typedef afw::image::MaskedImage<float>::Variance::Pixel VarPixel;
-    ndarray::EigenView<PsfPixel, 1, 1, Eigen::ArrayXpr> model(
-            fitRegion.getSpans()->flatten(psfImage->getArray(), psfImage->getXY0()));
-    ndarray::EigenView<float, 1, 1, Eigen::ArrayXpr> data(fitRegion.getSpans()->flatten(
-            exposure.getMaskedImage().getImage()->getArray(), exposure.getXY0()));
-    ndarray::EigenView<VarPixel, 1, 1, Eigen::ArrayXpr> variance(fitRegion.getSpans()->flatten(
-            exposure.getMaskedImage().getVariance()->getArray(), exposure.getXY0()));
+    auto model = fitRegion.getSpans()
+                         ->flatten(psfImage->getArray(), psfImage->getXY0())
+                         .asEigen<Eigen::ArrayXpr>();
```
kfindeisen: I don't understand why you need an explicit template argument here. What does it do, anyway? `asEigen` doesn't have any documentation...

PR #119, opened 2018-06-11
DM-14740: Stop using ndarray::EigenView indirectly in C++ code

r-owen:


src/PsfFlux.cc
```diff
@@ -93,22 +93,23 @@ void PsfFluxAlgorithm::measure(afw::table::SourceRecord& measRecord,
...
+    // while we use an Eigen::Map view of it
+    auto modelNdArray = fitRegion.getSpans()->flatten(psfImage->getArray(), psfImage->getXY0());
+    auto dataNdArray = fitRegion.getSpans()->flatten(exposure.getMaskedImage().getImage()->getArray(),
+                                                     exposure.getXY0());
+    auto varianceNdArray = fitRegion.getSpans()->flatten(exposure.getMaskedImage().getVariance()->getArray(),
+                                                         exposure.getXY0());
+    auto model = ndarray::asEigenMatrix(modelNdArray);
+    auto data = ndarray::asEigenMatrix(dataNdArray);
+    auto variance = ndarray::asEigenMatrix(varianceNdArray);
+    PsfPixel alpha = model.squaredNorm();
```
kfindeisen: Why are these `asEigenMatrix` rather than `asEigenArray`? The original code used `asEigen<Eigen::ArrayXpr>`.

r-owen: Because in almost all cases in the old code the array EigenView is used as a matrix by appending `.matrix()`. This eliminates all those calls to `.matrix()` at the expense of adding one call to `.array()`. See the changes below.

PR #123, opened 2018-07-16
DM-15126: Make meas_base compatible with pybind11 2.2 (as well as 2.1)

r-owen:


python/lsst/meas/base/algorithm.cc
```diff
@@ -31,6 +31,7 @@ using namespace pybind11::literals;
 namespace lsst {
 namespace meas {
 namespace base {
+namespace {
 
 PYBIND11_PLUGIN(algorithm) {
```
kfindeisen: This is different from the [examples in the style guide](https://developer.lsst.io/pybind11/style.html#wrapper-code-in-source-files-must-be-placed-in-a-nested-anonymous-namespace). Are we now required to put `PYBIND11_PLUGIN` in an anonymous namespace? I can't find any mention of anonymous namespaces in the pybind11 docs...

r-owen:
I thought using the anonymous namespace around everything was our standard, but I was wrong. I'll drop that change (while making sure that functions are still in anonymous).

python/lsst/meas/base/algorithm.cc
r-owen: I'm quite surprised. I thought our standard called for putting everything in anonymous namespace -- it's certainly easier that way, as then one can add functions as needed without worrying about it. But I have redone my changes to follow the style guide.

r-owen:
@kfindeisen based on your comments on slack I found and fixed the underlying error in `algorithms.cc`. This allowed the wrapper to work as before, eliminating the need to change the other two `meas_*` packages. I will drop those two ticket branches.

PR #128, opened 2018-08-23
 DM-10302 Change flux->instFlux

parejkoj:


include/lsst/meas/base/ApertureFlux.h
```diff
@@ -210,9 +210,9 @@ class ApertureFluxAlgorithm : public SimpleAlgorithm {
 };
 
 /**
- *  A Result struct for running an aperture flux algorithm with a single radius.
+ *  A Result struct for running an aperture instFlux algorithm with a single radius.
  *
- *  This simply extends FluxResult to add the appropriate error flags for aperture fluxes.
+ *  This simply extends FluxResult to add the appropriate error flags for aperture instFluxes.
```
r-owen: Is the flag for instFlux different than the flag for flux? If they share a flag then this change seems a bit off

include/lsst/meas/base/ApertureFlux.h
```diff
@@ -234,9 +234,9 @@ struct ApertureFluxResult : public FluxResult {
 };
 
 /**
- *  Measurement transformation for aperture fluxes
+ *  Measurement transformation for aperture instFluxes
  *
- *  Transforms fluxes with associated errors to magnitudes. Correctly handles
+ *  Transforms instFluxes with associated errors to magnitudes. Correctly handles
```
r-owen: Does this truly only change instFluxes or both kinds of fluxes?

include/lsst/meas/base/Blendedness.h
```diff
@@ -42,13 +42,14 @@ class BlendednessControl {
             doOld, bool,
             "Whether to compute HeavyFootprint dot products (the old deblend.blendedness parameter)");
 
-    LSST_CONTROL_FIELD(doFlux, bool, "Whether to compute quantities related to the Gaussian-weighted flux");
+    LSST_CONTROL_FIELD(doFlux, bool,
+                       "Whether to compute quantities related to the Gaussian-weighted instFlux");
```
r-owen: Is there any possibility that we'll want a flag that controls computation of non-inst Gaussian flux and/or non-Gaussian flux of either kind? If so, this name is worrisome. I hope there's no danger, as I suspect it's a nuisance to change.

include/lsst/meas/base/FluxUtilities.h
```diff
@@ -36,33 +36,35 @@ namespace meas {
...
- *  This class makes it easy to copy fluxes and their uncertainties to and from records, and provides
+ *  This class makes it easy to copy instFluxes and their uncertainties to and from records, and provides
  *  a method to add the appropriate fields to a Schema.
  */
 class FluxResultKey : public afw::table::FunctorKey<FluxResult> {
 public:
     /**
-     *  Add a pair of _flux, _fluxErr fields to a Schema, and return a FluxResultKey that points to them.
+     *  Add a pair of _instFlux, _instFluxErr fields to a Schema, and return a FluxResultKey that points to
+     * them.
```
r-owen: Again, I worry about the name: if this is FluxResultKey then what do you call the corresponding object for non-inst fluxes? InstFluxResultKey seems safer, but I have no idea how disruptive such a change would be.

include/lsst/meas/base/PeakLikelihoodFlux.h
```diff
@@ -38,16 +38,16 @@ namespace meas {
 namespace base {
 
 /**
- *  @brief C++ control object for peak likelihood flux.
+ *  @brief C++ control object for peak likelihood instFlux.
  *
- * Peak likelihood flux requires an image that has been filtered by convolving with its own PSF
- * (or an approximate model). It is equivalent to a PSF flux (e.g. PsfFluxAlgorithm) computed on
+ * Peak likelihood instFlux requires an image that has been filtered by convolving with its own PSF
+ * (or an approximate model). It is equivalent to a PSF instFlux (e.g. PsfFluxAlgorithm) computed on
```
r-owen: "It is equivalent to a PSF flux (e.g. PsfFluxAlgorithm)" is pre-existing but I find the "e.g." very confusing. I think the author meant "e.g. as computed by PsfFluxAlgorithm" and if agree and you are willing, please update accordingly. If not, or if in doubt, then please consider ditching the parenthetical comment.

include/lsst/meas/base/PeakLikelihoodFlux.h
```diff
@@ -38,16 +38,16 @@ namespace meas {
 namespace base {
 
 /**
- *  @brief C++ control object for peak likelihood flux.
+ *  @brief C++ control object for peak likelihood instFlux.
```
r-owen: Consider using "instrument flux" instead of "instFlux" here and in most of the rest of the documentation. I think it would enhance clarity.

python/lsst/meas/base/apCorrRegistry.py
```diff
@@ -20,29 +20,29 @@
...
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
-"""Registry for names of flux fields that should be aperture corrected
+"""Registry for names of instFlux fields that should be aperture corrected
 """
 
 __all__ = ("addApCorrName", "getApCorrNameSet")
 
-# Set of names of algorithms that measure fluxes that can be aperture corrected
+# Set of names of algorithms that measure instFluxes that can be aperture corrected
```
r-owen: Consider `instrument fluxes` instead of `instFluxes` here, and if you change t please also look for other instances and consider if they would also be clearer spelled out.

python/lsst/meas/base/applyApCorr.py
```diff
@@ -207,8 +207,8 @@ def run(self, catalog, apCorrMap):
                 source.set(apCorrInfo.apCorrFlagKey, True)
                 oldFluxFlagState = False
                 if self.config.doFlagApCorrFailures:
-                    oldFluxFlagState = source.get(apCorrInfo.fluxFlagKey)
-                    source.set(apCorrInfo.fluxFlagKey, True)
+                    oldFluxFlagState = source.get(apCorrInfo.instFluxFlagKey)
+                    source.set(apCorrInfo.instFluxFlagKey, True)
```
r-owen: I think you and Jim B wanted a single flag to be shared by both kinds of flux, so shouldn't this remain "fluxFlagKey"?

src/FluxUtilities.cc
```diff
@@ -30,28 +30,28 @@ namespace meas {
...
 
 FluxResultKey FluxResultKey::addFields(afw::table::Schema& schema, std::string const& name,
                                        std::string const& doc) {
     FluxResultKey result;
-    result._flux = schema.addField<Flux>(schema.join(name, "flux"), doc, "count");
-    result._fluxErr = schema.addField<FluxErrElement>(schema.join(name, "fluxErr"),
-                                                        "1-sigma flux uncertainty", "count");
+    result._instFlux = schema.addField<InstFlux>(schema.join(name, "instFlux"), doc, "count");
+    result._instFluxErr = schema.addField<InstFluxErrElement>(schema.join(name, "instFluxErr"),
+                                                          "1-sigma instFlux uncertainty", "count");
```
r-owen: Are you sure "Flux" and "FluxErrElement" need renaming? I would have naively expected that the same classes could be used as field types for both kinds of flux.

include/lsst/meas/base/ApertureFlux.h
parejkoj: Looking at the code, it is an instFlux->magnitude transform only. Though I'm a little curious why it even exists.

include/lsst/meas/base/FluxUtilities.h
parejkoj: If/when we eventually implement the new `_flux` fields, they could be attached to the FluxResultKey as `flux` and `fluxErr`, since they should share `flag` with `instFlux`.

PR #132, opened 2018-10-18
DM-16070: Implement RFC-534: Update naming of base_Blendedness fields

laurenam:


src/Blendedness.cc
```diff
@@ -197,41 +198,59 @@ BlendednessAlgorithm::BlendednessAlgorithm(Control const& ctrl, std::string cons
...
                 schema.join(name, "old"),
-                "blendedness from dot products: (child.dot(parent)/child.dot(child) - 1)");
+                "Blendedness from dot products: (child.dot(parent)/child.dot(child) - 1)");
     }
     if (_ctrl.doFlux) {
         _instFluxRaw = schema.addField<double>(
-                schema.join(name, "raw_instFlux"),
-                "measure of how instFlux is affected by neighbors: (1 - instFlux.child/instFlux.parent)");
+                schema.join(name, "raw"),
+                "Measure of how much instFlux is affected by neighbors: "
```
TallJimbo: For this docstring and that of just `abs`, I think I'd just say "flux" - because it's a flux ratio, it doesn't change for instrumental vs. calibrated, so it's weird to say "instrumental".

It would also be good to change the name of the variable `_instFluxRaw` to just `_raw`, too, so that isn't named inconsistently with the actual field.

src/Blendedness.cc
```diff
@@ -197,41 +198,59 @@ BlendednessAlgorithm::BlendednessAlgorithm(Control const& ctrl, std::string cons
...
+                "(1 - child_instFlux/parent_instFlux).  Operates on the \"raw\" pixel values.");
         _instFluxChildRaw = schema.addField<double>(
-                schema.join(name, "raw_instFlux_child"),
-                "instFlux of the child, measured with a Gaussian weight matched to the child", "count");
+                schema.join(name, "raw_child_instFlux"),
+                "instFlux of the child, measured with a Gaussian weight matched to the child.  "
```
TallJimbo: I'd spell out "Instrumental flux" instead of "instFlux" in the docstrings for this and other fields that end in `_instFlux`.

PR #135, opened 2019-01-29
DM-17474: Convert ForcedPhotCcd and ForcedPhotCoadd into pipelineTasks.

czwa:
The conversion of ForcedPhotCoadd was straightforward.  ForcedPhotCcd
required more handling, with duplicate operations in adjustArgsAndRun
to work around Gen-2 butler methods in the source filtering and
transformation code.

### DIFF
--- a/python/lsst/meas/base/forcedPhotCcd.py
+++ b/python/lsst/meas/base/forcedPhotCcd.py
@@ -145,6 +145,25 @@ class ForcedPhotCcdConfig(ForcedPhotImageConfig):
         default=False
     )
 
+    def setDefaults(self):
+        super().setDefaults()
+
+        # Override the Gen3 datasets from ForcedPhotImageTaskConfig.
+        # When these nameTemplate values are used in ForcedPhotCoadd,
+        # there is a coadd name that may change, depending on the
+        # coadd type.  For CCD forced photometry, there will likely
+        # only ever be a single calexp type, but for consistency, use
+        # the nameTemplate with nothing to substitute.
+        self.outputSchema.nameTemplate = "forced_src_schema"
+        self.exposure.nameTemplate = "{inputName}"
+        self.exposure.dimensions = ["Instrument", "Visit", "Detector"]
+        self.measCat.nameTemplate = "forced_src"
+        self.measCat.dimensions = ["Instrument", "Visit", "Detector", "SkyMap", "Tract"]
+
+        self.formatTemplateNames({"inputName": "calexp",
+                                  "inputCoaddName": "deep"})
+        self.quantum.dimensions = ("Instrument", "Visit", "Detector", "SkyMap", "Tract")
+
 
 class ForcedPhotCcdTask(ForcedPhotImageTask):
     """A command-line driver for performing forced measurement on CCD images.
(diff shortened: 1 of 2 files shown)


python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -111,8 +163,30 @@ class ForcedPhotImageTask(lsst.pipe.base.CmdLineTask):
...
+            idFactory = lsst.afw.table.IdFactory.makeSource(expId, 64 - expBits)
+
+            inputData['measCat'] = self.measurement.generateMeasCat(inputData['exposure'],
+                                                                    inputData['refCat'],
+                                                                    inputData['refWcs'],
+                                                                    idFactory=idFactory)
+
+        return super().adaptArgsAndRun(inputData, inputDataIds, outputDataIds, butler)
+
+    def __init__(self, butler=None, refSchema=None, initInputs=None, **kwds):
```
yalsayyad: I'm used to the `__init__` method going first.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -182,10 +234,72 @@ class ForcedPhotCcdTask(ForcedPhotImageTask):
...
+
+        # Step 4: Transform source footprints from the reference
+        # coordinates to the exposure coordinates.
+        for srcRecord, refRecord in zip(sources, refSources):
+            srcRecord.setFootprint(refRecord.getFootprint().transform(inputData['refWcs'],
+                                                                      expWcs, expRegion))
+        # Step 5: Replace reference catalog with filtered source list.
+        inputData['refCat'] = sources
+
+        if 'measCat' not in inputData.keys():
```
yalsayyad: Would you add a comment saying why you're adding `measCat` to the inputs when its an output?  When do we expect it to be in  `inputData.keys()`?

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -39,9 +39,55 @@
...
+    exposure = pipeBase.InputDatasetField(
+        doc="Input exposure to perform photometry on.",
+        name="",
+        nameTemplate="{inputCoaddName}Coadd",
+        scalar=True,
+        storageClass="ExposureF",
+        dimensions=["AbstractFilter", "SkyMap", "Tract", "Patch"],
+    )
+    refCat = pipeBase.InputDatasetField(
+        doc="Reference catalog of sources.",
```
yalsayyad: Catalog of shapes and positions at which to force photometry

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -111,8 +163,30 @@ class ForcedPhotImageTask(lsst.pipe.base.CmdLineTask):
     ConfigClass = ForcedPhotImageConfig
     _DefaultName = "processImageForcedTask"
 
-    def __init__(self, butler=None, refSchema=None, **kwds):
-        super(lsst.pipe.base.CmdLineTask, self).__init__(**kwds)
+    def getInitOutputDatasets(self):
+        return {"outputSchema": afwTable.SourceCatalog(self.measurement.schema)}
+
+    def adaptArgsAndRun(self, inputData, inputDataIds, outputDataIds, butler):
+        if 'measCat' not in inputData.keys():
```
yalsayyad: There's some code duplication in this block. It looks identical but with "TractPatch" instead of "VisitDetector"

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -25,10 +25,10 @@
 `ForcedPhotCcdTask`, `ForcedPhotCoaddTask`).
 """
 
-import lsst.afw.table
+import lsst.afw.table as afwTable
```
yalsayyad: Ditto on https://developer.lsst.io/python/style.html#import-statements-should-follow-a-standard-pattern

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -182,10 +234,72 @@ class ForcedPhotCcdTask(ForcedPhotImageTask):
...
 
     ConfigClass = ForcedPhotCcdConfig
-    RunnerClass = lsst.pipe.base.ButlerInitializedTaskRunner
+    RunnerClass = pipeBase.ButlerInitializedTaskRunner
     _DefaultName = "forcedPhotCcd"
     dataPrefix = ""
 
+    def adaptArgsAndRun(self, inputData, inputDataIds, outputDataIds, butler):
+        # Filtering the reference catalog is currently handled by
+        # Gen2 specific methods.  To get around this, port in code
```
yalsayyad: We've been trying to keep `adaptArgsAndRun` cleaner than `runDataRef`. Could you move these 3 steps into a method called `filterReferences()` with a docstring explaining that the old `fetchReferences()` did some filtering in addition to the fetching? (If that accurately reflects why this is necessary of course)

yalsayyad: And if the code is duplicated elsewhere, a note in the docstring about where to look would help future devs.

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -111,8 +163,30 @@ class ForcedPhotImageTask(lsst.pipe.base.CmdLineTask):
     ConfigClass = ForcedPhotImageConfig
     _DefaultName = "processImageForcedTask"
 
-    def __init__(self, butler=None, refSchema=None, **kwds):
-        super(lsst.pipe.base.CmdLineTask, self).__init__(**kwds)
+    def getInitOutputDatasets(self):
+        return {"outputSchema": afwTable.SourceCatalog(self.measurement.schema)}
```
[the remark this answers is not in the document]
natelust: InitOutputs dict is not set except by end code that returns some data. I think you might be thinking of getInitOutputaDatasetTypes

python/lsst/meas/base/forcedPhotImage.py
czwa: Moved into new method.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -139,12 +139,64 @@ def imageOverlapsTract(tract, imageWcs, imageBox):
 
 
 class ForcedPhotCcdConfig(ForcedPhotImageConfig):
+    # Gen 3 options
+    inputSchema = pipeBase.InitInputDatasetField(
+        doc="Schema for the input measurement catalogs.",
+        name="",
```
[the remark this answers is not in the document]
czwa: This matches my memory (that the example I worked from had both), but I've removed the empty assignments, and will test things are consistent before merging.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -139,12 +139,64 @@ def imageOverlapsTract(tract, imageWcs, imageBox):
...
+        nameTemplate="{inputCoaddName}Coadd_ref_schema",
+        storageClass="SourceCatalog",
+    )
+    outputSchema = pipeBase.InitOutputDatasetField(
+        doc="Schema for the output forced measurement catalogs.",
+        name="",
+        nameTemplate="forced_src_schema",
```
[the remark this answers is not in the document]
czwa: Converting this to a name, and using the forcedPhotImage defaults doesn't work, which matches with nameTemplate overriding the names previously.  I've left this as a nameTemplate, which works correctly.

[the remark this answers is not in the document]
TallJimbo: Agree that this works, but I think it merits a code comment (which would now go in `setDefaults`, I think).

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -145,6 +145,19 @@ class ForcedPhotCcdConfig(ForcedPhotImageConfig):
...
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.outputSchema.nameTemplate = "forced_src_schema"
+        self.exposure.nameTemplate = "{inputName}"
+        self.exposure.dimensions = ["Instrument", "Visit", "Detector"]
+        self.measCat.nameTemplate = "forced_src"
+        self.measCat.dimensions = ["Instrument", "Visit", "Detector", "SkyMap", "Tract"]
+
+        self.formatTemplateNames({"inputName": "calexp",
```
TallJimbo: I think this may have gotten out of sync with other changes (or maybe I'm just commenting on a branch you're still working on): I don't see "inputName" in the substitution strings in the config anymore.

TallJimbo: Oh, nevermind, I see how it works now.  Probably merits a code comment, though.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -186,6 +199,90 @@ class ForcedPhotCcdTask(ForcedPhotImageTask):
...
+                        break
+                if topId == 0:
+                    sources.append(record)
+                    refSources.append(record)
+
+        # Step 4: Transform source footprints from the reference
+        # coordinates to the exposure coordinates.
+        for srcRecord, refRecord in zip(sources, refSources):
+            srcRecord.setFootprint(refRecord.getFootprint().transform(refWcs,
+                                                                      expWcs, expRegion))
```
TallJimbo: I don't see why we need two catalogs containing the same records; couldn't you just do:
```
for refRecord in refSources:
    refRecord.setFootprint(refRecord.getFootprint().transform(refWcs, expWcs, expRegion))
```
?

python/lsst/meas/base/forcedPhotImage.py
```diff
@@ -123,6 +174,53 @@ def __init__(self, butler=None, refSchema=None, **kwds):
             self.makeSubtask("applyApCorr", schema=self.measurement.schema)
         self.makeSubtask('catalogCalculation', schema=self.measurement.schema)
 
+    def getInitOutputDatasets(self):
+        return {"outputSchema": lsst.afw.table.SourceCatalog(self.measurement.schema)}
+
+    def adaptArgsAndRun(self, inputData, inputDataIds, outputDataIds, butler):
+        # ForcedPhotImageTask.run() expects to work in-place on measCat. This requires it to be
+        # constructed here before being passed as an input to run(). It should not be in the
+        # inputData dict(), but check that we don't clobber something.
```
TallJimbo: If it shouldn't be there but is, that should probably raise an exception.

czwa: Running last minute tests shows I needed to be a bit clearer in the comment (for myself if nothing else).  If measCat doesn't exist, it should be made, and for the ForcedPhotCoadd case, which just calls ForcedPhotImageTask, we do not expect it to exist.  However, for ForcedPhotCcd, we construct the measCat there, and then pass it up to the parent class in adaptArgsAndRun.  In this case, it does exist, and shouldn't be made.

PR #139, opened 2019-02-15
Tickets/DM-17757 Fix creating imports when saving config

natelust:
Change default module to None 

Changing to None causes the module to be looked up from the
control class, as numerical lookup proved to be unreliable.

### DIFF
--- a/python/lsst/meas/base/wrappers.py
+++ b/python/lsst/meas/base/wrappers.py
@@ -68,7 +68,7 @@ class WrappedForcedPlugin(ForcedPlugin):
         self.cpp.fail(measRecord, error.cpp if error is not None else None)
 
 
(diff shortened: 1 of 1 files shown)


python/lsst/meas/base/wrappers.py
```diff
@@ -68,7 +68,7 @@ def fail(self, measRecord, error=None):
         self.cpp.fail(measRecord, error.cpp if error is not None else None)
 
 
-def wrapAlgorithmControl(Base, Control, module=2, hasMeasureN=False):
+def wrapAlgorithmControl(Base, Control, module=None, hasMeasureN=False):
```
TallJimbo: Should update the parameter docs to reflect the new default as well.

PR #140, opened 2019-02-21
DM-15514: Write insertFakeObjectsTask to add fake sources for QA purposes

sr525:


natelust:
The commit message conveys almost no information, try a different title and or include more information in the commit message body

python/lsst/meas/base/forcedPhotCoadd.py
```diff
@@ -39,6 +39,8 @@ class ForcedPhotCoaddConfig(ForcedPhotImageConfig):
         optional=True
     )
 
+    fakesAdded = lsst.pex.config.Field(dtype=bool, default=False, doc="Have fakes been added?")
```
natelust: This Field should follow the structure of the other fields already in the file, aka multi-line

Documentation should be statements and not questions, perhapse something like "Fields indicates if Fake objects are present in input dataset"

Consider a different name for the field variable, like fakesPresent of fakesInserted as it is closer to the use of this variable in English terms, but this is just a suggestion from someone reading it, there is nothing wrong with how it is

python/lsst/meas/base/forcedPhotCoadd.py
```diff
@@ -88,7 +90,12 @@ class ForcedPhotCoaddTask(ForcedPhotImageTask):
     dataPrefix = "deepCoadd_"
 
     def getExposure(self, dataRef):
-        name = self.config.coaddName + "Coadd_calexp"
+
+        if self.config.fakesAdded:
+            name = "fakes_" + self.config.coaddName + "Coadd_calexp"
```
natelust: In general the python community seems to be going toward using the format method or f strings instead of the addition operator, so it might be useful to get in the habit. This code would then be
```
name = f"fakes_{self.config.coaddName}Coadd_calex"
```
What you have will run fine, just using the opportunity to maybe pass along some tips. There is a runtime performance you gain by using f strings or format vs concatenation, but in this case the micro optimization does not really matter

sr525: Is it preferable to use this way or to have a consistent formatting throughout the file?

python/lsst/meas/base/forcedPhotCoadd.py
sr525: I've changed the structure of the Field but have let the name. I have used fakesAdded in my examples for the DESC meeting and would prefer to keep things consistent.

python/lsst/meas/base/forcedPhotCoadd.py
laurenam: https://developer.lsst.io/coding/intro.html?highlight=consistency#deviating-from-the-dm-style-guides (second bullet):

> Coding consistency is very important but sometimes the style guide doesn’t apply either due to lack of a definitive rule or the circumstances of the specific code segment logically dictate otherwise. When in doubt, use your best judgment or ask the lead developer.
> 
> Here are two plausible reasons to break a particular rule:
> - When applying the rule would make the code less readable, even for someone who is used to reading code that follows the rules.
> - To be consistent with surrounding code that also breaks it (maybe for historic reasons)—although this is also an opportunity to clean up someone else’s mess.

PR #145, opened 2019-06-14
DM-12594: Remove daf::base::Citizen

parejkoj:


tests/test_SincPhotSums.py
```diff
@@ -221,11 +221,15 @@ def tearDown(self):
 
     def assertCached(self, coeff1, coeff2):
         np.testing.assert_array_equal(coeff1.getArray(), coeff2.getArray())
-        self.assertEqual(coeff1.getId(), coeff2.getId())
+        # This compares the memory address and read-only attributes of the images.
+        self.assertEqual(coeff1.array.__array_interface__['data'],
+                         coeff2.array.__array_interface__['data'])
```
natelust: While I am filled with a little discomfort using internal array interfaces, I don't really see a good way to test what you want to test without it. the only change I might suggest is using something like
`coeff1.array.ctypes.data` as it reads a little clearer, and is attribute access, so if numpy decides to change something they can make the attribute access return the same thing, but dict access may or may not need to change. I dont feel too strongly about this though.

parejkoj: Good catch, I didn't see the ctypes version of that. I like it better too.

PR #150, opened 2019-10-15
DM-5635: not flagged NaN in sdssCentroid

jdswinbank:


isullivan:
This looks good, though I wonder if the exception should be raised in `_centroidChecker`:
https://github.com/lsst/meas_base/blob/5d04eb760b412ea8c4487f38f6fd620ca66d6ac9/src/CentroidUtilities.cc#L200
That looks to me like that would be the correct place to check the centroid values, but I don't know if that would have consequences for other uses of that function.

jdswinbank:
It's a good thought, and, in fact, I had the same one earlier.

I didn't implement this in `CentroidChecker`  for a few reasons:

- As you suggest, it might have an impact on other users of that code.
- The implementation would be a little awkward, as not all centroiders provide errors, so you'd need to check whether `xErr`/`yErr` exist, as well as checking them for NaNs.
- The interface is a little awkward, as CentroidChecker adds new flags to the algorithm output which aren't defined in the algorithm code itself. That makes them harder to document, and harder for users to understand where they've come from.

On the other hand:

- Grep tells me there are few other users of the code, so it would be easy to handle this.
- The precedent for the ugly interface has already been set, so maybe this wouldn't make it much worse.
- It would be easier to test, since `CentroidChecker` is a much less convoluted piece of code.

Given the above, I jumped the way I did, but if you'd prefer to add this check to `CentroidChecker` I'd be happy to shift it there — I don't have very strong opinions myself.

isullivan:
Those are all good points, though I think adding this one check here makes the existing ugly interface worse because it splits up where the flags are set. I think I would have a mild preference for moving it.

jdswinbank:
Well, note that _almost all_ flags are set by throwing `MeasurementError` in algorithmic code. In this case, `edge`, `noSecondDerivative`, `almostNoSecondDerivative`, and `notAtMaximum` are all set in that way; `resetToPeak` is the only one that isn't.

But as I say, I don't feel strongly about this, and I would like to test this better, so I'm happy to move it. Watch this space. :-)

jdswinbank:
@isullivan — any better?

include/lsst/meas/base/CentroidUtilities.h
```diff
@@ -190,15 +190,20 @@ class CentroidTransform : public BaseTransform {
 class CentroidChecker {
 public:
     /**
-     *  Check source record for an centroid algorithm called name, noting if the centroid
-     *  already set by the algorithm is outside the footprint attached to the record.
-     *  If that is the case, we should:
-     *     (1) set the general failure flag for the algorithm
-     *     (2) set the algorithmName + "_flag_resetToPeak" flag
-     *     (3) change the value of the centroid to the footprint Peak
+     *  Check source record produced by a centroid algorithm called name.
```
isullivan: Please put quotes around "name".

include/lsst/meas/base/CentroidUtilities.h
```diff
@@ -190,15 +190,20 @@ class CentroidTransform : public BaseTransform {
...
      *
-     *  Secondly, check to see if the centroid is more than "dist" pixels from
-     *  the footprint peak, and similarly modify the centroi value to the peak in that case
+     *  If the centroid is accompanied uncertainties (the xErr and yErr
```
isullivan: "is accompanied uncertainties" seems to be missing a word. Maybe "by"?

include/lsst/meas/base/CentroidUtilities.h
```diff
@@ -190,15 +190,20 @@ class CentroidTransform : public BaseTransform {
...
+     *  fields), these should not be NaN. If they are, the algorithmName +
+     *  "_flag_badError" flag and the general failure flag for the algorithm
+     *  are both set.
+     *
+     *  If the centroid set by the algorithm outside the footprint attached to
```
isullivan: If the centroid is set..., or ...is outside...?

PR #152, opened 2020-01-29
DM-20841: Support for approximating WCS as SIP

PaulPrice:


erykoff:
Oh yes, thank you for noticing.  I think we need to be able to specify external photometry / wcs separately as well, going from doUberCal to doApplyExternal... .  I fear this should have been part of DM-21308, or at least it's closely related.  If this is too much for this ticket, you can fill another and assign it to me.

python/lsst/meas/base/readCalibratedExposure.py
```diff
@@ -0,0 +1,203 @@
...
+                Photometric calibration (`lsst.afw.image.PhotoCalib`), or
+                ``None`` if not ``doApplyExternalPhotoCalib``.
+            ``skyWcs``
+                Astrometric calibration (`lsst.afw.geom.SkyWcs`), or ``None``
+                if not doApplyExternalSkyWcs``.
+        """
+        photoCalib = None
+        if self.config.doApplyExternalPhotoCalib:
+            photoCalib = dataRef.get(f"{self.config.externalPhotoCalibName}_photoCalib")
+
```
erykoff: I think this needs to send back the exposure's {{photoCalib}} even if we aren't applying an external one, or else we don't actually scale the exposure properly below.

PR #153, opened 2020-02-07
DM-23196: Add calibration storage to multiBand.py

morriscb:


python/lsst/meas/base/plugins.py
```diff
@@ -400,6 +400,134 @@ def fail(self, measRecord, error=None):
...
+    """
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return BasePlugin.FLUX_ORDER
+
+    def __init__(self, config, name, schema, metadata):
+        GenericPlugin.__init__(self, config, name, schema, metadata)
+        self.photoKey = schema.addField(
+            "base_LocalPhotoCalib",
```
yalsayyad: Use `name` to reuse the name you specify when calling makeSingleFramePlugin Below.

python/lsst/meas/base/plugins.py
```diff
@@ -400,6 +400,134 @@ def fail(self, measRecord, error=None):
...
+
+
+class EvaluateLocalPhotoCalibPlugin(GenericPlugin):
+    """Evaluate the local value of the Photometric Calibration in the exposure.
+
+    The aim is to store the local calib value within the catalog for later
+    use in the Science Data Model functors.
+    """
+
+    ConfigClass = VarianceConfig
```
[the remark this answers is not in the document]
morriscb: Already removed.

python/lsst/meas/base/plugins.py
morriscb: So you're saying that in the class `__init__` I should use the `name` variable in to replace explicitly writing `base_LocalPhotoCalib` or `base_LocalWcs`?

PR #154, opened 2020-02-11
DM-23352: Add jointcal, skyCorr to forcedPhotCcd

PaulPrice:


python/lsst/meas/base/recalibrateExposure.py
```diff
@@ -0,0 +1,321 @@
...
+            )
+
+        class FooTaskConnections(RecalibrateExposureConnections):
+            ...
+
+        class FooTask(lsst.pipe.base.PipelineTask):
+            def runQuantum(self, butlerQC, inputRefs, outputRefs):
+                inputs = butlerQC.get(inputRefs)
+                calibs = self.recalibrate.extractCalibs(inputs, len(inputs["exposure"]))
+                self.recalibrate.runMultiple(inputs["exposure"], **calibs)
```
TallJimbo: Instead of providing `runMultiple`, I think I'd recommend that `runQuantum` implementors just call `run` as many times as they need, on the explicit exposure-like datasets they actually have.  Making `runQuantum` less magic at the expense of a bit of boilerplate is a net win, especially because most of the time there will only be a single call to `run`.

python/lsst/meas/base/recalibrateExposure.py
```diff
@@ -0,0 +1,321 @@
...
+                outputs = self.run(**inputs)
+                ...
+    """
+    ConfigClass = RecalibrateExposureConfig
+
+    def extractCalibs(self, inputs, number=None, single=False):
```
TallJimbo: Using "Calibs" in this name (and in the return value name, the local variable in the example above, and `readCalibs`) isn't wrong, but we have such a strong association of that word with the low-level ISR master calibrations that I think it's worth looking for an alternative.  "Recalibrations" might work; where the subtask name already provides some context, `extractDatasets` and `readDatasets` might as well.

python/lsst/meas/base/recalibrateExposure.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Returns
+        -------
+        exposure : `lsst.afw.image.Exposure`
+            Recalibrated exposure.
+        """
+        if photoCalib is None:
+            photoCalib = exposure.getPhotoCalib()
+        exposure.maskedImage = photoCalib.calibrateImage(exposure.maskedImage,
+                                                         self.config.includePhotoCalibVar)
```
TallJimbo: I think this step should be optional - it's expensive, and we don't need it if we're only going to do catalog operations (e.g. `ForcedPhotCcd`!), for which attaching a spatially-varying `PhotoCalib` to the `Exposure` is sufficient.

python/lsst/meas/base/recalibrateExposure.py
```diff
@@ -0,0 +1,321 @@
...
+        if rescale:
+            scaling = photoCalib.getCalibrationMean()
+            exposure.maskedImage /= scaling
+            exposure.setPhotoCalib(PhotoCalib(scaling, photoCalib.getCalibrationErr()))
+        else:
+            exposure.setPhotoCalib(PhotoCalib(1.0))  # calibrateImage makes 1 ADU = 1 nJy everywhere
```
TallJimbo: Rather that putting it in an `else` block here, I'd move this just below the `calibrateImage` call, because it's critical that this line happens if that one does - we don't want to have any chance the exposure's `PhotoCalib` could remain set to the spatially-varying one, and while the logic here is totally correct, separating those lines by the `if rescale` block leaves that invariant vulnerable to future refactors.

natelust:
I don't have much to add on top of what Jim has said, except to point out that you do in fact need the inheritance here for the connections. We can look at alternative ways to do things going forward, but what you did there is the right thing in this case.

PR #155, opened 2020-03-24
DM-23946: Change localWcs plugins to use a localGnomonicWcs transform.

TallJimbo:
Debug unittest and plugin.

### DIFF
--- a/python/lsst/meas/base/plugins.py
+++ b/python/lsst/meas/base/plugins.py
@@ -467,6 +467,7 @@ class EvaluateLocalWcsPlugin(GenericPlugin):
     use in the Science Data Model functors.
     """
     ConfigClass = EvaluateLocalWcsPluginConfig
+    _scale = (1.0 * lsst.geom.arcseconds).asDegrees()
 
     @classmethod
     def getExecutionOrder(cls):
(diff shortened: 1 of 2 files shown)


python/lsst/meas/base/plugins.py
```diff
@@ -467,6 +467,7 @@ class EvaluateLocalWcsPlugin(GenericPlugin):
     use in the Science Data Model functors.
     """
     ConfigClass = EvaluateLocalWcsPluginConfig
+    _scale = (1.0 * lsst.geom.arcseconds).asDegrees()
```
TallJimbo, line 470: Better to make this all-caps, since it's being used as a constant.

python/lsst/meas/base/plugins.py
```diff
@@ -475,32 +476,64 @@ def getExecutionOrder(cls):
     def __init__(self, config, name, schema, metadata):
         GenericPlugin.__init__(self, config, name, schema, metadata)
         self.cdMatrix11Key = schema.addField(
-            "%s_CDMatrix_1_1" % name,
+            f"{name}_CDMatrix_1_1",
```
TallJimbo, line 479: Overall I'm not a fan of using "CD matrix" as the name for these (just "matrix" or "jacobian" would be better), as that implies a FITS WCS meaning that's subtly incorrect (related to the fact that just getting the CD matrix of the WCS _doesn't_ do we want).  But feel free to leave it for now if changing it would be disruptive in a way that would expand the scope of the ticket.

python/lsst/meas/base/plugins.py
```diff
@@ -475,32 +476,64 @@ def getExecutionOrder(cls):
...
+        measRecord.set(self.cdMatrix22Key, localMatrix[1, 1])
+
+    def makeLocalTransformMatrix(self, wcs, center):
+        """Create a local, linear approximation of the wcs transformation
+        matrix.
+
+        The approximation is created as if the center is at RA=0, DEC=0. All
+        comparing x,y coordinate are relative to the position of center. Matrix
+        is initially calculated with units arcseconds and then converted to
+        radians. This yields higher precision results due to quirks in AST.
```
TallJimbo, line 514: I'd move the note about using arcseconds initially to a code comment; a user of the function shouldn't care.

timj: It's not entirely clear to me how this can be an AST problem as such given that AST doesn't use arcseconds itself at all and does everything in radians.

TallJimbo: My best theory was the AST might be using the output unit of the mapping to inform the step size for finite differences, which would be reasonable (and in some contexts probably desirable) behavior even if it otherwise uses radians internally.  But it was only ever a theory, and I think the important point here is just that arcseconds work best here and we don't really care that we don't fully understand why.

PR #162, opened 2020-08-17
DM-25768: Remove computeFluxScale function

taranu:


src/SdssShape.cc
```diff
@@ -871,8 +857,8 @@ FluxResult SdssShapeAlgorithm::computeFixedMomentsFlux(ImageT const &image,
                                       .str());
         }
         double var = ImageAdaptor<ImageT>().getVariance(image, ix, iy);
-        double i0Err = std::sqrt(var / wArea);
-        result.instFluxErr = i0Err * 2 * wArea;
+        // i0Err = std::sqrt(var / wArea); instFluxErr = 2 * wArea * i0Err
```
isullivan, line 860: Remove the commented-out code, but add a real comment to show where your final formula came from.

src/SdssShape.cc
```diff
@@ -796,8 +776,10 @@ SdssShapeResult SdssShapeAlgorithm::computeAdaptiveMoments(ImageT const &image,
...
     }
-    if (result.getQuadrupole().getIxx() * result.getQuadrupole().getIyy() <
-        (1.0 + 1.0e-6) * result.getQuadrupole().getIxy() * result.getQuadrupole().getIxy())
+    double IxxIyy = result.getQuadrupole().getIxx() * result.getQuadrupole().getIyy();
+    double Ixy_sq = result.getQuadrupole().getIxy();
+    Ixy_sq *= Ixy_sq;
+    if (IxxIyy < (1.0 + 1.0e-6) * Ixy_sq)
     // We are checking that Ixx*Iyy > (1 + epsilon)*Ixy*Ixy where epsilon is suitably small. The
     // value of epsilon used here is a magic number. DM-5801 is supposed to figure out if we are
     // to keep this value.
```
isullivan, line 785: If you can think of a better error message than the one below, please do so. But, it is outside the scope of this ticket, so don't feel you have to. If I got that error while processing data, I'd have to search the code for the message in order to figure out what went wrong.

PR #165, opened 2020-12-02
Use sum instead of amplitude to calculate flux

arunkannawadi:


src/SdssShape.cc
```diff
@@ -839,15 +841,13 @@ FluxResult SdssShapeAlgorithm::computeFixedMomentsFlux(ImageT const &image,
     double const w22 = std::get<3>(weights);
     bool const interp = shouldInterp(shape.getIxx(), shape.getIyy(), std::get<0>(weights).second);
 
-    double i0 = 0;  // amplitude of Gaussian
+    double sum0 = 0;  //  sum of pixel values weighted by a Gaussian
     if (calcmom<true>(ImageAdaptor<ImageT>().getImage(image), localCenter.getX(), localCenter.getY(), bbox,
-                      0.0, interp, w11, w12, w22, &i0, NULL, NULL, NULL, NULL, NULL, NULL, NULL) < 0) {
+                      0.0, interp, w11, w12, w22, NULL, &sum0, NULL, NULL, NULL, NULL, NULL, NULL) < 0) {
```
TallJimbo, line 846: If you're saying here that `sum0` is now _not_ the amplitude of the Gaussian, but the weighted sum over pixel values, doesn't that mean the docs for `calcmom` above are (and always were) wrong?  Or is this some limit where they are in fact the same?

arunkannawadi: No, I am not saying anything was wrong earlier. `calcmom` can take pointers for both the amplitude and the sum, depending on the position of the argument. So I supply a NULL pointer for amplitude and a `double` pointer for the sum.

TallJimbo: Ah, sorry, I missed the fact that the argument changed placers here.

TallJimbo (approved):

PR #170, opened 2021-03-03
 DM-29023 Remove traces of ForcedPhotImage

natelust:


czwa (approved):
I think this catches all the cases I can see.

PR #171, opened 2021-03-30
DM-29501: replace unary_function, binary_function, deprecated in C++17

mwittgen:


src/SincCoeffs.cc
```diff
@@ -146,7 +146,7 @@ class CircularAperture {
 };
 
 template <typename CoordT>
-class CircApPolar : public std::unary_function<CoordT, CoordT> {
+class CircApPolar {
```
erykoff, line 149: I couldn't find any uses of this code by doing a search, but I guess it doesn't hurt to leave it around?  (Same for `SincAperture` and `GaussPowerFunctor` below).

erykoff (approved):

PR #172, opened 2021-04-26
DM-29767: Take absolute value of deviation to compare with uncertainty measures

arunkannawadi:
In a few unit tests, the signed deviation is checked against a positive number and the assertion is True if the deviation is lesser than the estimated uncertainty. The assertion would pass for large negative deviations, which the unit test is supposed to catch. The fixes in this PR take absolute value of the deviations before comparing to the estimated uncertainties.


cmsaunders (approved):
Everything looks good!

arunkannawadi:
Thanks, Clare!

PR #174, opened 2021-05-07
DM-29936: getChildren without repeatedly checking if the SourceCatalog is sorted

arunkannawadi:


python/lsst/meas/base/forcedMeasurement.py
```diff
@@ -353,10 +353,11 @@ def run(self, measCat, exposure, refCat, refWcs, exposureId=None, beginOrder=Non
         # Create parent cat which slices both the refCat and measCat (sources)
         # first, get the reference and source records which have no parent
         refParentCat, measParentCat = refCat.getChildren(0, measCat)
-        for parentIdx, (refParentRecord, measParentRecord) in enumerate(zip(refParentCat, measParentCat)):
-
+        childrenIter = refCat.getChildren((refParentRecord.getId() for refParentRecord in refCat), measCat)
+        for parentIdx, records in enumerate(zip(refParentCat, measParentCat, childrenIter)):
+            # Unpack records
+            (refParentRecord, measParentRecord, (refChildCat, measChildCat)) = records
```
fred3m, line 359: This is being very nit-picky, but you don't need the outer parentheses on the left hand side.

fred3m (approved):
Everything looks good. I left one very nit-picky comment that you can ignore if you like.

python/lsst/meas/base/forcedMeasurement.py
arunkannawadi: Thanks for pointing that out. I had those parenthesis initially because it was in the `for` statement which led to a ridiculously long line. I should remove the outer parenthesis according to the dev_guide [here](https://developer.lsst.io/python/style.html?highlight=python#the-minimum-number-of-parentheses-needed-for-correctness-and-readability-should-be-used).

PR #177, opened 2021-05-26
DM-30046: Defer reading refcats in forcedPhotCcd

MorganSchmitz:
[Jira ticket](https://jira.lsstcorp.org/browse/DM-30046)

Jenkins run [#34305](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/34305/pipeline)

### DIFF
--- a/python/lsst/meas/base/forcedPhotCcd.py
+++ b/python/lsst/meas/base/forcedPhotCcd.py
@@ -172,7 +172,8 @@ class ForcedPhotCcdConnections(PipelineTaskConnections,
         name="{inputCoaddName}Coadd_ref",
         storageClass="SourceCatalog",
         dimensions=["skymap", "tract", "patch"],
-        multiple=True
+        multiple=True,
+        deferLoad=True,
     )
     skyMap = cT.Input(
         doc="SkyMap dataset that defines the coordinate system of the reference catalog.",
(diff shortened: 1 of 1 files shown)


python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -395,14 +397,19 @@ def mergeAndFilterReferences(self, exposure, refCats, refWcs):
...
+            refCat = refCat.get()
+            if mergedRefCat is None:
+                mergedRefCat = lsst.afw.table.SourceCatalog(refCat.table)
             containedIds = {0}  # zero as a parent ID means "this is a parent"
             for record in refCat:
                 if expPolygon.contains(record.getCoord().getVector()) and record.getParent() in containedIds:
                     record.setFootprint(record.getFootprint().transform(refWcs, expWcs, expRegion))
                     mergedRefCat.append(record)
                     containedIds.add(record.getId())
+        if mergedRefCat is None:
```
arunkannawadi, line 411: Question more than a comment - will we get here only if refCats is an empty list? Would it ever be the case where refCats have empty tables, in which case this might still have to be raised? Sure, it wasn't raised earlier, but just wondering.

arunkannawadi (approved):

python/lsst/meas/base/forcedPhotCcd.py
MorganSchmitz: Ah, reading it I think you're right, if all records are empty we wouldn't raise. I'm never quite sure what happens with empty things in general, and the best to handle it in particular. From what I gathered at discussions following DM-29862, I'm not sure there really _is_ a standard way of dealing with such cases yet.

As you say, we never raised before anyway; this addition was just Jim figuring we might as well throw a clearer exception since we needed to change the way we handled `mergedRefCat`. I'm going to go ahead and merge - I'm sure we can live with postponing handling the case of empty-only records until after this weekly gets tagged :)

Thanks for the instantaneous review, Arun!

arunkannawadi: If `refCats` was empty, then line 398 before would have thrown an `IndexError`. Now, we make it throw a `RuntimeError`. I'm betting this doesn't make any difference, unless the file calling this was catching specific exceptions in which case that would have to be changed as well. Just wanted to point that out here.

PR #178, opened 2021-06-16
DM-30386: Make DRP wrapper task for DiaCalculation task

morriscb:


kherner (commented):
Actually this comment was supposed to be here: One question on the deleted diaCalculation.py and diaCalculationPlugins.py files. When I compare the lines in the "new" versions in meas_base I see there are 514 and 1,110 lines, respectively, but 463 and 1,093 in the now-deleted versions. The test files are the same though. Do you know off the top of your head where the difference come from? I cna do a detailed comparison but if this expected maybe we can save some time.

morriscb:
Here's the response I left on the other pull request:

It's due to 1) the extra for loop over plugins outside of the one that loops over filters first. 2) adding the needsFilter boolean to each plugin. 3) try/except clauses to check if the filter exists in the dataset.

kherner (approved):

PR #179, opened 2021-07-09
DM-31060: Clean up logging usage

timj:


czwa (approved):
This all looks consistent with the description on the ticket and RFC-782.

PR #181, opened 2021-07-16
DM-29346: Compute PSFlux chi squared metric

parejkoj:


include/lsst/meas/base/PsfFlux.h
```diff
@@ -90,6 +90,8 @@ class PsfFluxAlgorithm : public SimpleAlgorithm {
     Control _ctrl;
     FluxResultKey _instFluxResultKey;
     afw::table::Key<float> _areaKey;
+    afw::table::Key<float> _chi2Key;
+    afw::table::Key<int> _npixelsKey;
```
kfindeisen, line 94: What's the difference between `_npixelsKey` and `_areaKey`? Consider adding (short) comments to these fields; the only context I have is that this is a `PsfFluxAlgorithm`.

src/PsfFlux.cc
```diff
@@ -110,10 +113,15 @@ void PsfFluxAlgorithm::measure(afw::table::SourceRecord& measRecord,
...
     measRecord.set(_areaKey, model.sum() / alpha);
+    measRecord.set(_npixelsKey, fitRegion.getSpans()->getArea());
     if (!std::isfinite(result.instFlux) || !std::isfinite(result.instFluxErr)) {
         throw LSST_EXCEPT(PixelValueError, "Invalid pixel value detected in image.");
     }
     measRecord.set(_instFluxResultKey, result);
+    auto chi2 = ((data.cast<PsfPixel>() - result.instFlux * model).array().square() *
+                 variance.cast<PsfPixel>().array().rsqrt())
+                        .sum();
+    measRecord.set(_chi2Key, chi2);
```
kfindeisen, line 124: While I'm not a particular fan of reduced χ^2 (AFAIK statisticians don't take it or "degrees of freedom" seriously), I don't see where the 1/N factor you promised comes from?

tests/test_PsfFlux.py
```diff
@@ -32,6 +32,22 @@
                                   SingleFramePluginTransformSetupHelper)
 
 
+def compute_chi2(exposure, centroid, instFlux, maskPlane=None):
+    """Return the chi2 for the exposure PSF at the given position."""
```
kfindeisen, line 36: More documentation? E.g., it's not obvious right away what `maskPlane` is, and you don't mention returning the pixel count.

tests/test_PsfFlux.py
```diff
@@ -32,6 +32,22 @@
...
+    psfImage = exposure.getPsf().computeImage(centroid)
+    scaledPsf = psfImage.array*instFlux
+    # Get a sub-image the same size as the returned PSF image.
+    sub = exposure.Factory(exposure, psfImage.getBBox(), lsst.afw.image.LOCAL)
+    if maskPlane is not None:
+        masked = np.logical_not(sub.mask.array & sub.mask.getPlaneBitMask(maskPlane))
```
kfindeisen, line 42: Shouldn't this be called `unmasked` or `nonmasked`? Or, per the [style guide](https://developer.lsst.io/cpp/style.html#negated-boolean-variable-names-must-be-avoided), don't include the `not` in the variable definition?

tests/test_PsfFlux.py
```diff
@@ -32,6 +32,22 @@
...
+        chi2 = np.sum((sub.image.array[masked] - scaledPsf[masked])**2 / np.sqrt(sub.variance.array[masked]))
+        nPixels = masked.sum()
+    else:
+        chi2 = np.sum((sub.image.array - scaledPsf)**2 / np.sqrt(sub.variance.array))
```
kfindeisen, line 46: Again, is there a 1/N missing?

tests/test_PsfFlux.py
```diff
@@ -100,11 +121,20 @@ def testSubImage(self):
...
         subExposure = exposure.Factory(exposure, bbox, lsst.afw.image.LOCAL)
+
         algorithm.measure(record, subExposure)
         self.assertFloatsAlmostEqual(record.get("base_PsfFlux_instFlux"), record.get("truth_instFlux"),
                                      atol=3*record.get("base_PsfFlux_instFluxErr"))
         self.assertTrue(record.get("base_PsfFlux_flag_edge"))
 
+        # Expected chi2 requires trimming PSF image by one pixel per side
+        # to match subExposure created above, so we can't use compute_chi2() directly here
+        scaledPsf = psfImage.array[1:-1, 1:-1]*record.get("base_PsfFlux_instFlux")
```
kfindeisen, line 132: Why do you need to trim the image, and why can't `compute_chi2` be designed to handle this case?

tests/test_PsfFlux.py
```diff
@@ -144,11 +180,19 @@ def testMonteCarlo(self):
...
+                measuredNPixels.append(record.get("base_PsfFlux_npixels"))
+                chi2, nPixels = compute_chi2(exposure, record.getCentroid(), instFluxes[-1])
+                expectedChi2s.append(chi2)
+                expectedNPixels = nPixels  # should be the same each time
             instFluxMean = np.mean(instFluxes)
             instFluxErrMean = np.mean(instFluxErrs)
             instFluxStandardDeviation = np.std(instFluxes)
             self.assertFloatsAlmostEqual(instFluxErrMean, instFluxStandardDeviation, rtol=0.10)
             self.assertLess(abs(instFluxMean - instFlux), 2.0*instFluxErrMean / nSamples**0.5)
+            self.assertFloatsAlmostEqual(np.mean(measuredChi2s), np.mean(expectedChi2s), rtol=1e-7)
```
kfindeisen, line 193: I don't understand why you're comparing the means here. Shouldn't `measuredChi2s[i] == expectedChi2s[i]` for all `i`?

kfindeisen (changes requested):
I have some questions about the tests, but the main issue is that the code claims it computes reduced χ^2 but seems to actually be computing χ^2 itself (in both the algorithm and the test). This discrepancy needs to be reconciled one way or the other.

include/lsst/meas/base/PsfFlux.h
parejkoj: These keys get doc fields assigned in the `PsfFluxAlgorithm` constructor; I'll copy those docs to comments here.

src/PsfFlux.cc
parejkoj: The 1/N is a lie, and I've removed it from the comments/docs.

kfindeisen (approved):

tests/test_PsfFlux.py
parejkoj: I've renamed it `unmasked`: I think in this case that problem described in the style guide shouldn't be an issue.

tests/test_PsfFlux.py
parejkoj: Since `compute_chi2()` here is only to support these tests, I didn't want to figure out how to make it support this other use case.

The comment just above this line doesn't help you with the "why" question?

tests/test_PsfFlux.py
parejkoj: No, that's only true for the number of pixels. The chi2 is different for each individual sample at a given noise level because the data+noise is randomly generated.

tests/test_PsfFlux.py
kfindeisen: Not really. You don't normally need to match anything, do you?

parejkoj: The C++ code handles matching the PSF image and exposure image dimensions automatically. I didn't write the python code above to do the same, and just copied the relevant bits here. I tweaked the comment, not sure if it helps:

```
        # Calculating chi2 requires trimming the PSF image by one pixel per side
        # to match the subExposure created above, so we can't use compute_chi2()
        # directly here.
```

tests/test_PsfFlux.py
parejkoj: Oh! I get your comment now. That's absolutely right: I've changed it to do the comparison per-sample.

src/PsfFlux.cc
```diff
@@ -110,10 +113,15 @@ void PsfFluxAlgorithm::measure(afw::table::SourceRecord& measRecord,
     // variance as if we had, so we'll apply the weights to the model now, and update alpha.
     result.instFluxErr = std::sqrt(model.array().square().matrix().dot(variance.cast<PsfPixel>())) / alpha;
     measRecord.set(_areaKey, model.sum() / alpha);
+    measRecord.set(_npixelsKey, fitRegion.getSpans()->getArea());
     if (!std::isfinite(result.instFlux) || !std::isfinite(result.instFluxErr)) {
         throw LSST_EXCEPT(PixelValueError, "Invalid pixel value detected in image.");
     }
     measRecord.set(_instFluxResultKey, result);
+    auto chi2 = ((data.cast<PsfPixel>() - result.instFlux * model).array().square() *
+                 variance.cast<PsfPixel>().array().rsqrt())
```
ebellm, line 122: This isn't right: we want the variance (sigma squared) in the denominator.

```suggestion
                 variance.cast<PsfPixel>().array())
```

tests/test_PsfFlux.py
```diff
@@ -32,6 +32,40 @@
...
+        The computed chi2 and number of pixels included in the calculation.
+    """
+    psfImage = exposure.getPsf().computeImage(centroid)
+    scaledPsf = psfImage.array*instFlux
+    # Get a sub-image the same size as the returned PSF image.
+    sub = exposure.Factory(exposure, psfImage.getBBox(), lsst.afw.image.LOCAL)
+    if maskPlane is not None:
+        unmasked = np.logical_not(sub.mask.array & sub.mask.getPlaneBitMask(maskPlane))
+        chi2 = np.sum((sub.image.array[unmasked] - scaledPsf[unmasked])**2
+                      / np.sqrt(sub.variance.array[unmasked]))
```
ebellm, line 61: again, no square root in the denominator.

tests/test_PsfFlux.py
```diff
@@ -32,6 +32,40 @@
...
+        nPixels = unmasked.sum()
+    else:
+        chi2 = np.sum((sub.image.array - scaledPsf)**2 / np.sqrt(sub.variance.array))
```
ebellm, line 64: no square root

ebellm (changes requested):
There is a math error, indeed--see comments throughout.

src/PsfFlux.cc
kfindeisen: D'oh! I'm embarrassed I missed that. 🤦

parejkoj:
Good catch, thank you @ebellm ! I'll rerun my gen3_rc2_subset test and upload a new plot on the ticket.

PR #183, opened 2021-08-18
DM-31442: Pass log to undeblended plugins

arunkannawadi:
This is very similar to the changes made in DM-430. The plugins have different call signatures depending on whether they do logging internally or not. This was not accounted for 'undeblended' plugins.

### DIFF
--- a/python/lsst/meas/base/baseMeasurement.py
+++ b/python/lsst/meas/base/baseMeasurement.py
@@ -269,7 +269,7 @@ class BaseMeasurementTask(lsst.pipe.base.Task):
         for executionOrder, name, config, PluginClass in sorted(self.config.plugins.apply()):
             #   Pass logName to the plugin if the plugin is marked as using it
             #   The task will use this name to log plugin errors, regardless.
-            if hasattr(PluginClass, "hasLogName") and PluginClass.hasLogName:
+            if getattr(PluginClass, "hasLogName", False):
                 self.plugins[name] = PluginClass(config, name, metadata=self.algMetadata,
                                                  logName=self.log.getChild(name).name, **kwds)
             else:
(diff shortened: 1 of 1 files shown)


python/lsst/meas/base/baseMeasurement.py
```diff
@@ -280,8 +280,13 @@ def initializePlugins(self, **kwds):
         # Initialize the plugins to run on the undeblended image
         for executionOrder, name, config, PluginClass in sorted(self.config.undeblended.apply()):
             undeblendedName = self.config.undeblendedPrefix + name
-            self.undeblendedPlugins[name] = PluginClass(config, undeblendedName, metadata=self.algMetadata,
-                                                        **kwds)
+            if hasattr(PluginClass, "hasLogName") and PluginClass.hasLogName:
```
timj, line 284: Looking at this now, I don't like the duplication of the complex arguments below and I don't like the code this was copied from above. I think I would much prefer it to be something like:

```python
logName = None
if hasattr(PluginClass, "hasLogName") and PluginClass.hasLogName:
    logName = self.log.getChild(name).name
self.undeblendedPlugins[name] = PluginClass(config, undeblendedName, metadata=self.algMetadata,
                                            logName=logName, **kwds)
```
Then it is clear that the `if`-statement is solely modifying the `logName` parameter and it doesn't force you to look at the other parameters to ensure they haven't been changed in the two cases. Please fix the code above in the same way.

timj (changes requested):

python/lsst/meas/base/baseMeasurement.py
timj: Actually, I think this could be even cleaner:
```python
if getattr(PluginClass, "hasLogName", False):
    logName = self.log.getChild(name).name
self.undeblendedPlugins[name] = PluginClass(config, undeblendedName, metadata=self.algMetadata,
                                            logName=logName, **kwds)
```
Maybe I've missed something obvious though since I'm a little tired at the moment.

arunkannawadi: I think this logic is completely taken care in `wrappers.py` and we might be able to always pass in a `logName` unconditionally. Let me try that out tomorrow.

arunkannawadi: Hey Tim, this won't work because not all plugins have `logName` as one of their arguments. So there will have to be two different calls, depending on whether the `__init__` of the `PluginClass` can take `logName` as a keyword argument or not. We can certainly simplify the logic though. Or a more painful exercise is to make all the plugins in our codebase to accept it as an optional argument.

timj (approved):
Thanks for cleaning this up.

PR #184, opened 2021-08-31
DM-31528: Add more log messages to the measure task

arunkannawadi:
The changes in the PR ensures that logging happens (at VERBOSE level) at least every 10 minutes (or at an interval that is configurable) when measurement is run.


timj (approved):
Looks good. Same comment about when to add the 600 seconds as I made in lsst/meas_extensions_scarlet#43

PR #186, opened 2021-09-09
DM-28394: Evaluate local photocalib and Wcs during ForcedPhotCcd

yalsayyad:


python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -275,6 +275,7 @@ def setDefaults(self):
         # ForcedMeasurementTask
         super().setDefaults()
 
+        self.measurement.plugins.names |= ['base_LocalPhotoCalib', 'base_LocalWcs']
```
cmsaunders, line 278: Does this work? If `self.measurement.plugins.names` is a list, I don't think it will.

cmsaunders (commented):
I'm not sure that this will work as written...

python/lsst/meas/base/forcedPhotCcd.py
yalsayyad: This is a "add some plugin names to plugin.names" idiom we use a lot: e.g. https://github.com/lsst/obs_subaru/blob/master/config/cmodel.py https://github.com/lsst/obs_subaru/blob/master/config/kron.py

But looks like DM-31601 is going to be merged first, so this getting rebased off of this ticket.

TallJimbo: To explain a bit: `.names` is actually a `set`-like object.  But it also has implicit conversions from `list`, because it dates back to before `set` literals existed.

Now that I think about it, that might make
```
config.measurement.names |= {"plugin1", "plugin2"}
```
a clearer choice for new code.

cmsaunders (approved):

PR #187, opened 2021-09-12
DM-31601: Add Tasks to force photometry from DataFrames

yalsayyad:
Tasks seed forced photometry with ra/decl columns from
DataFrames instead of SourceCatalogs.

### DIFF
--- a/python/lsst/meas/base/forcedPhotCcd.py
+++ b/python/lsst/meas/base/forcedPhotCcd.py
@@ -21,6 +21,8 @@
 
 import collections
 import logging
+import pandas as pd
+import numpy as np
 
 import lsst.pex.config
 import lsst.pex.exceptions
@@ -47,7 +49,8 @@ try:
 except ImportError:
     applyMosaicResults = None
 
-__all__ = ("PerTractCcdDataIdContainer", "ForcedPhotCcdConfig", "ForcedPhotCcdTask", "imageOverlapsTract")
+__all__ = ("PerTractCcdDataIdContainer", "ForcedPhotCcdConfig", "ForcedPhotCcdTask", "imageOverlapsTract",
+           "ForcedPhotCcdFromDataFrameTask", "ForcedPhotCcdFromDataFrameConfig")
 
 
 class PerTractCcdDataIdContainer(pipeBase.DataIdContainer):
@@ -274,7 +277,7 @@ class ForcedPhotCcdConfig(pipeBase.PipelineTaskConfig,
         # Make catalogCalculation a no-op by default as no modelFlux is setup by default in
         # ForcedMeasurementTask
         super().setDefaults()
-
+        self.measurement.plugins.names |= ['base_LocalPhotoCalib', 'base_LocalWcs']
         self.catalogCalculation.plugins.names = []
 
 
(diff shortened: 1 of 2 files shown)


python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -682,3 +685,141 @@ def _makeArgumentParser(cls):
...
+    outputSchema = cT.InitOutput(
+        doc="Schema for the output forced measurement catalogs.",
+        name="forced_src_diaObject_schema",
+        storageClass="SourceCatalog",
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        self.initInputs.remove("inputSchema")
+        self.inputs.remove("skyMap")
```
TallJimbo, line 717: This is a sign that it'd be better to make a new ABC for both concrete implementations, instead of inheriting from the other concrete impl.  I'd say that's a job for another ticket, but these are going to still show up in the `config.connections` for this task, and it can be a real pain to remove configs with breakage.  Some copy-and-paste duplication would probably be better.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -682,3 +685,141 @@ def _makeArgumentParser(cls):
...
+        self.makeSubtask("measurement", refSchema=lsst.afw.table.SourceTable.makeMinimalSchema())
+
+        if self.config.doApCorr:
+            self.makeSubtask("applyApCorr", schema=self.measurement.schema)
+        self.makeSubtask('catalogCalculation', schema=self.measurement.schema)
+        self.outputSchema = lsst.afw.table.SourceCatalog(self.measurement.schema)
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        print([i.dataId for i in inputs['refCat']])
```
TallJimbo, line 756: Debug relic.

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -682,3 +685,141 @@ def _makeArgumentParser(cls):
...
+                                                                       inputs['exposure'], inputs['refCat'],
+                                                                       inputs['refWcs'],
+                                                                       "visit_detector")
+        outputs = self.run(**inputs)
+        butlerQC.put(outputs, outputRefs)
+
+    def df2RefCat(self, dfList, exposureBBox, exposureWcs):
+        """Convert list of DataFrames to reference catalog
+
+        Concatenate list of DataFrames presumable from multiple patches and
```
TallJimbo, line 772: "presumable" -> "presumably"?

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -682,3 +685,141 @@ def _makeArgumentParser(cls):
...
+        dfList : `list` of `pandas.DataFrame`
+            Each element containst diaObjects with ra/decl position in degrees
+            Columns 'diaObjectId', 'ra', 'decl' are expected
+        exposureBBox :   `lsst.geom.Box2I`
+            Bounding box on which to select rows that overlap
+        exposureWcs : `lsst.afw.geom.SkyWcs`
+            World coordinate system to convert sky coords in ref cat to
+            pixel coords with which to compare with exposureBBox
+
+
```
TallJimbo, line 786:
```suggestion
```

python/lsst/meas/base/plugins.py
```diff
@@ -766,6 +768,47 @@ def measure(self, measRecord, exposure, refRecord, refWcs):
...
+    """
+    pass
+
+
+@register("base_TransformedCentroidFromCoord")
+class ForcedTransformedCentroidFromCoordPlugin(ForcedTransformedCentroidPlugin):
+    """Record the transformation of the reference catalog coord.
+    The coord recorded in the reference catalog is tranformed to the
+    measurement coordinate system and stored.
+    Parameters
```
TallJimbo, line 782:
```suggestion

    Parameters
```

python/lsst/meas/base/plugins.py
```diff
@@ -766,6 +768,47 @@ def measure(self, measRecord, exposure, refRecord, refWcs):
...
+    config : `ForcedTransformedCentroidFromCoordConfig`
+        Plugin configuration
+    name : `str`
+        Plugin name
+    schemaMapper : `lsst.afw.table.SchemaMapper`
+        A mapping from reference catalog fields to output
+        catalog fields. Output fields are added to the output schema.
+    metadata : `lsst.daf.base.PropertySet`
+        Plugin metadata that will be attached to the output catalog.
+    Notes
```
TallJimbo, line 793:
```suggestion

    Notes
```

python/lsst/meas/base/plugins.py
```diff
@@ -766,6 +768,47 @@ def measure(self, measRecord, exposure, refRecord, refWcs):
...
+    -----
+    This can be used as the slot centroid in forced measurement when only a
+    reference coord exits, allowing subsequent measurements to simply refer to
```
TallJimbo, line 796:
```suggestion
    reference coord exists, allowing subsequent measurements to simply refer to
```

TallJimbo (approved):

python/lsst/meas/base/forcedPhotCcd.py
yalsayyad: Thanks, what I'm reading is that:
1) First choice is an ABC 
2) Second choice is a non-inherited Connections class.

TallJimbo: Yup, exactly.

TallJimbo: ...and I'm not super confident that (1) is even better than (2); I'm just guessing that is because you chose to inherit in the first place.

PR #188, opened 2021-09-13
DM-30927: Set calexp WCS to None for failed astrometric fit

laurenam:


erykoff (approved):
I've never seen so many force pushes! 😭

PR #191, opened 2021-09-22
DM-31869: Add sensible positions to computeShape in tests

arunkannawadi:
The `GaussianPsf` objects are position-independent and are not affected
by lack of position argument. However, with RFC-796 in mind, sensible
positions are passed.


erykoff (approved):
Looks good!

PR #192, opened 2021-09-27
DM-31849: Fix miscellaneous forced photometry bugs.

TallJimbo:


natelust (approved):
A few minor comments. It all seems fine, but I make no guarantees about subtitle issues, all this forced stuff takes a lot to load into memory and grok. I support a rewrite soon.

PR #194, opened 2021-10-03
DM-8146: fix fftw failure when using MKL based conda stack

mwittgen:


src/SincCoeffs.cc
```diff
@@ -447,11 +449,11 @@ std::shared_ptr<afw::image::Image<PixelT>> calcImageKSpaceReal(double const rad1
             double const airy1 = (rad1 > 0 ? rad1 * J1(twoPiRad1 * k) : 0.0) / k;
             double const airy2 = rad2 * J1(twoPiRad2 * k) / k;
             double const airy = airy2 - airy1;
-            c[fY * wid + fX] = airy;
+            c[fY * wid + fX] = {airy, 0.};
```
rmjarvis, line 452: I don't think this is required.  IMO, this and the one below are more readable having the rhs just be the real value.  But YMMV.

PR #197, opened 2021-10-18
DM-32252: Add variance_median and improve test coverage.

erykoff:


tests/test_NoiseReplacer.py
```diff
@@ -75,15 +76,76 @@ def setUp(self):
...
+        task.run(catalog, exposure)
+        sumVariance = exposure.getMaskedImage().getVariance().getArray().sum()
+        for record in catalog:
+            self.assertFloatsAlmostEqual(record.get("test_NoiseReplacer_inside"),
+                                         record.get("truth_instFlux"), rtol=1E-3)
+
+            # N.B. Next line checks that a random value is correct to a
+            # statistical 1-sigma prediction; some RNG seeds may cause it to
+            # fail (indeed, 67% should)
+            self.assertLess(record.get("test_NoiseReplacer_outside"), np.sqrt(sumVariance))
```
mfisherlevine, line 116: Isn't this just an exact copy and paste of the test above, meaning you could do the whole thing with
```
for noiseSource in ['measure', 'variance']:
    task.config.noiseReplacer.noiseSource = noiseSource```

File: tests/test_NoiseReplacer.py [hunk 2]
```diff
@@ -75,15 +76,76 @@ def setUp(self):
...
+        exposure.setMetadata(md)
+        task.run(catalog, exposure)
+        sumVariance = exposure.getMaskedImage().getVariance().getArray().sum()
+        for record in catalog:
+            self.assertFloatsAlmostEqual(record.get("test_NoiseReplacer_inside"),
+                                         record.get("truth_instFlux"), rtol=1E-3)
+
+            # N.B. Next line checks that a random value is correct to a
+            # statistical 1-sigma prediction; some RNG seeds may cause it to
+            # fail (indeed, 67% should)
```
mfisherlevine, line 139: OK, so this comment claims that 67% should fail statistically, but there's actually been no fine tuning there - all the random seeds are zero, and all apparently pass. What's up with that? (Yes, it could be luck, but still, worth wondering here).

PR #199, opened 2021-10-20
DM-32285: Fix yy <-> xy bug in SdssShape uncertainties

arunkannawadi:
Due to a convention mismatch, the `_yyErr` and `_xyErr` components are interchanged for `base_SdssShape` plugin. This PR fixes this bug and adds a MonteCarlo unit test that would have caught it in the first place.

### DIFF
--- a/tests/test_SdssShape.py
+++ b/tests/test_SdssShape.py
@@ -25,6 +25,7 @@ import numpy as np
 
 import lsst.geom
 import lsst.afw.geom
+import lsst.afw.table
 import lsst.meas.base
 import lsst.meas.base.tests
 import lsst.utils.tests
@@ -47,6 +48,15 @@ class SdssShapeTestCase(lsst.meas.base.tests.AlgorithmTestCase, lsst.utils.tests
         del self.bbox
         del self.dataset
         del self.config
+
+    def makeAlgorithm(self, ctrl=None):
+        """Construct an algorithm and return both it and its schema.
+        """
+        if ctrl is None:
+            ctrl = lsst.meas.base.SdssShapeControl()
+        schema = lsst.meas.base.tests.TestDataset.makeMinimalSchema()
+        algorithm = lsst.meas.base.SdssShapeAlgorithm(ctrl, "base_SdssShape", schema)
+        return algorithm, schema
 
     def assertFinite(self, value):
         self.assertTrue(np.isfinite(value), msg="%s is not finite" % (value,))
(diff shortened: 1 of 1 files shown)


src/SdssShape.cc
```diff
@@ -597,18 +597,16 @@ bool getAdaptiveMoments(ImageT const &mimage, double bkgd, double xcen, double y
                 if (!(shape->flags[SdssShapeAlgorithm::UNWEIGHTED.number])) {
                     Matrix4d fisher = calc_fisher(*shape, bkgd_var);  // Fisher matrix
                     Matrix4d cov = fisher.inverse();
-                    // convention in afw::geom::ellipses is to order moments (xx, yy, xy),
-                    // but the older algorithmic code uses (xx, xy, yy) - the order of
-                    // indices here is not a bug.
+                    // convention is to order moments (xx, yy, xy)
```
jmeyers314, line 600: Looks good, but can you smash the first two commits so there isn't a commit with the code change but not the doc change?

tests/test_SdssShape.py
```diff
@@ -129,6 +139,47 @@ def testMeasureBadPsf(self):
             self._checkShape(result, record)
             self.assertTrue(result.getFlag(lsst.meas.base.SdssShapeAlgorithm.PSF_SHAPE_BAD.number))
 
+    def testMonteCarlo(self):
+        """Test an ideal simulation, with no noise.
```
jmeyers314, line 143: "with no noise" doesn't seem right.  Maybe "with known noise" or "with controlled noise"?

tests/test_SdssShape.py
```diff
@@ -129,6 +139,47 @@ def testMeasureBadPsf(self):
...
+
+        Demonstrate that:
+
+        - We get exactly the right answer, and
```
jmeyers314, line 147: I'd omit "exactly"

tests/test_SdssShape.py
```diff
@@ -129,6 +139,47 @@ def testMeasureBadPsf(self):
...
+                exposure, cat = self.dataset.realize(noise*instFlux, schema, randomSeed=i)
+                record = cat[1]
+                algorithm.measure(record, exposure)
+                catalog.append(record)
+
+            catalog = catalog.copy(deep=True)
+            for suffix in ["xx", "yy", "xy"]:
+                shapeMean = np.mean(catalog["base_SdssShape_"+suffix])
+                shapeErrMean = np.nanmean(catalog["base_SdssShape_"+suffix+"Err"])
+                shapeStandardDeviation = 0.5*(np.nanpercentile(catalog["base_SdssShape_"+suffix], 84)
```
jmeyers314, line 176: Maybe call this `shapeInterval68`?  I think it's only equivalent to the standard deviation for a Gaussian distribution.

tests/test_SdssShape.py
arunkannawadi: This was just copy-pasted from other MonteCarlo tests. But you're right that this is not accurate and will change.

src/SdssShape.cc
arunkannawadi: Fair point

PR #206, opened 2022-03-18
DM-33982: Add finalized psf and external calibration support to forced ccd photometry.

erykoff:


python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -360,20 +448,116 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
-                                                         inputs['refWcs'])
-
-        inputs['measCat'], inputs['exposureId'] = self.generateMeasCat(inputRefs.exposure.dataId,
-                                                                       inputs['exposure'],
-                                                                       inputs['refCat'], inputs['refWcs'],
-                                                                       "visit_detector")
-        self.attachFootprints(inputs['measCat'], inputs['refCat'], inputs['exposure'], inputs['refWcs'])
-        # TODO: apply external calibrations (DM-17062)
-        outputs = self.run(**inputs)
+        refWcs = skyMap[tract].getWcs()
```
yalsayyad, line 451: Only complaint with this runQuantim rewrite, that you changed more than you had to (no lines needed modifying at all), which unnecessarily rejects the `runQuantum` pattern of doing all the disk-touching and input modification (and literally modifying the inputs dict) and then passing the modified `**inputs` to run. You could just *add* these lines between the existing `inputs['refWcs'] = skyMap[tract].getWcs()` and `inputs['refCat'] = self.mergeAndFilterReferences(...`

```   
        # Connections only exist if they are configured to be used.
        skyCorr = inputs.get("skyCorr", None)
        if self.config.useGlobalExternalSkyWcs:
            externalSkyWcsCatalog = inputs.get("externalSkyWcsGlobalCatalog", None)
        else:
            externalSkyWcsCatalog = inputs.get("externalSkyWcsTractCatalog", None)
        if self.config.useGlobalExternalPhotoCalib:
            externalPhotoCalibCatalog = inputs.get("externalPhotoCalibGlobalCatalog", None)
        else:
            externalPhotoCalibCatalog = inputs.get("externalPhotoCalibTractCatalog", None)
        finalizedPsfApCorrCatalog = inputs.get("finalizedPsfApCorrCatalog", None)

        inputs['exposure'] = self.prepareCalibratedExposure(
            inputs["exposure"],
            skyCorr=skyCorr,
            externalSkyWcsCatalog=externalSkyWcsCatalog,
            externalPhotoCalibCatalog=externalPhotoCalibCatalog,
            finalizedPsfApCorrCatalog=finalizedPsfApCorrCatalog
        )
```

yalsayyad (approved):

python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -778,18 +1037,43 @@ def __init__(self, butler=None, refSchema=None, initInputs=None, **kwds):
 
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
+
+        tract = butlerQC.quantum.dataId['tract']
+        skyMap = inputs.pop('skyMap')
+        inputs['refWcs'] = skyMap[tract].getWcs()
```
yalsayyad, line 1043: So sorry for not noticing this the first round. 

This is more wrong than what was here before. The `refWcs` is not the tract WCS and adding `skymap` as an input to this task is unnecessary.  This Task only runs on data frames with ra/decl coords. No WCS needed  nor allowed.  See validation message in `self.validate`.  

Remove your change, and below where it used to say:
```
inputs['refWcs'] = inputs['exposure'].getWcs()
```

you're welcome to change it to:
```
inputs['refWcs'] = None
```

which is probably clearer than what was there before.

erykoff: After wading through the code, I see now that `generateMeasCat` takes in a `refWcs` argument that is never used by anything at all (???), and `attachFootprints` only uses `refWcs` when used in `transformed` mode which isn't allowed for this task.  So I'll change it to `None` which is indeed clearer than what was there before.

PR #208, opened 2022-04-04
DM-30163: Replace asserts by exceptions, cleanup calcmom

mwittgen:


arunkannawadi (commented):
Left a comment on Jira about how the exception should ideally be an InvalidParameterError and not a RuntimeError. Also, the commit message must start with a capitalized letter.

src/SdssShape.cc
```diff
@@ -560,8 +574,9 @@ bool getAdaptiveMoments(ImageT const &mimage, double bkgd, double xcen, double y
      */
     if (shape->flags[SdssShapeAlgorithm::UNWEIGHTED.number]) {
         w11 = w22 = w12 = 0;
-        if (calcmom<false>(image, xcen, ycen, bbox, bkgd, interpflag, w11, w12, w22, &I0, &sum, &sumx, &sumy,
-                           &sumxx, &sumxy, &sumyy, NULL, negative) < 0 ||
+        double ignored;
```
arunkannawadi, line 577: Curious why this is `ignored` in place of `sums4`.

src/SdssShape.cc
```diff
@@ -190,34 +190,23 @@ geom::Box2I computeAdaptiveMomentsBBox(geom::Box2I const &bbox,      // full ima
 /*
  * Calculate weighted moments of an object up to 2nd order
  */
-template <bool instFluxOnly, typename ImageT>
```
arunkannawadi, line 193: Why is the option to compute `instFluxOnly` removed? It may not be used regularly, but it seems like a perfectly good option to have.

arunkannawadi (approved):
I can confirm that the code that used to crash is now exiting gracefully with a useful error message. The commit has to split up into a clean up commit and the one that prevents the crash. My earlier comment about the capitalization was for the commit message, not the PR title. See https://developer.lsst.io/work/flow.html#appendix-commit-message-best-practices

import lsst.afw.geom as afwGeom
--
import lsst.afw.image as afwImage
import lsst.meas.base as measBase
import numpy as np
 
img = afwImage.ImageD(np.random.randn(16,16))
shape = afwGeom.Quadrupole(-2, -2, 0)  # <-- This is invalid
flux = measBase.SdssShapeAlgorithm.computeFixedMomentsFlux(img, shape,
img.getBBox().getCenter())

src/SdssShape.cc
mwittgen: calcmon is overloaded now instFluxOnly instantiated two different versions of the functions with C style pointers.One could rename calcmon to make this explicit.

src/SdssShape.cc
mwittgen: In the old code NULL was passed. One could rename it to sums4, but this is never used. 
The old code used pointers to pass optional parameters C-style.

src/SdssShape.cc
arunkannawadi: Ah, yes. Makes sense. I'm happy as long as there's an option to calculate only the `instFlux`.

mwittgen: Yeah, the old code was C translated into bad C++.

src/SdssShape.cc
arunkannawadi: Hmm, may be still keep the variable name to `psums4` or `sums4` but just add a comment saying it is ignored? It'll prevent looking up the definition in the future to see what that NULL corresponds to.

PR #212, opened 2022-05-28
DM-35008: Check whether the mask is present before asking for it

timj:
In some architectures the try/except catching a C++ exception
can be extremely slow and some code paths can call this
code many times in a loop. Replacing with the existence check
is significantly faster.


TallJimbo (approved):
I also think `addMaskPlane` does nothing but return the existing plane if the given one already exists, so even this check is unnecessary.  But it's up to you whether to try that and endure another round of testing.

PR #213, opened 2022-06-17
DM-35265: Reduce usage of MeasurementError

timj:


src/PsfFlux.cc
```diff
@@ -93,7 +93,9 @@ void PsfFluxAlgorithm::measure(afw::table::SourceRecord& measRecord,
                                    ->clippedTo(exposure.getMaskedImage().getMask()->getBBox()));
     }
     if (fitRegion.getArea() == 0) {
-        throw LSST_EXCEPT(MeasurementError, NO_GOOD_PIXELS.doc, NO_GOOD_PIXELS.number);
+        _flagHandler.setValue(measRecord, NO_GOOD_PIXELS.number, true);
+        _flagHandler.setValue(measRecord, FAILURE.number, true);
+        return;
```
timj, line 98: Shouldn't line 81 above also return? It looks like it sets the FAILURE flag and so is indicating that there is a real problem yet it is allowed to continue.

TallJimbo: Do do have a policy of algorithms attempting to produce the best measurement possible even when they know something is very wrong and set a flag to say so.  I suspect that's what's going on here.

src/SdssCentroid.cc
```diff
@@ -381,8 +367,7 @@ std::pair<MaskedImageT, double> smoothAndBinImage(std::shared_ptr<afw::detection
     try {
         subImage.reset(new MaskedImageT(mimage, bbox, afw::image::LOCAL));
     } catch (pex::exceptions::LengthError &err) {
-        throw LSST_EXCEPT(MeasurementError, SdssCentroidAlgorithm::EDGE.doc,
-                          SdssCentroidAlgorithm::EDGE.number);
+        return std::make_tuple(mimage, 0, SdssCentroidAlgorithm::EDGE.number);
```
TallJimbo, line 370: If you want to remove this `try/catch` as well, I think the replacement is:
```
if (!mimage.getBBox(afw::image::LOCAL).contains(bbox)) {
    return std::make_tuple(mimage, 0, SdssCentroidAlgorithm::EDGE.number);
}
subImage.reset(new MaskedImageT(mimage, bbox, afw::image::LOCAL));
```

TallJimbo (approved):
Looks good!  Only comment is a suggestion for how to get rid of the remaining `try/catch` on `LengthError`.

PR #221, opened 2022-08-05
DM-971: fix errors now that tasks always validate on init

parejkoj:


python/lsst/meas/base/baseMeasurement.py
```diff
@@ -158,6 +167,10 @@ class BaseMeasurementConfig(lsst.pex.config.Config):
     where ``PluginBaseClass`` is the appropriate base class of the plugin
     (e.g., `SingleFramePlugin` or `ForcedPlugin`).
     """
+    def __new__(cls, *args, ignoreSlotPluginChecks=False, **kwargs):
+        instance = super().__new__(cls, *args, **kwargs)
```
erykoff, line 171: If this is deprecated we should raise a FutureWarning if this is set to True.

python/lsst/meas/base/baseMeasurement.py
```diff
@@ -138,6 +138,15 @@ def setupSchema(self, schema):
...
     """Base configuration for all measurement driver tasks.
 
+    Parameters
+    ----------
+    ignoreSlotPluginChecks : `bool`, optional
+        Do not check that all slots have an associated plugin to run when
+        validating this config. This is primarily for tests that were written
+        before we made Tasks always call `config.validate()` on init.
+        DEPRECATED: this is a temporary workaround while we better define how
+        config/schema validation works for measurement tasks.
```
erykoff, line 148: There should be a TODO ticket number associated with this comment.

python/lsst/meas/base/baseMeasurement.py
erykoff: If you are worried about tests spewing out deprecation warnings, I wouldn't be opposed to wrapping them in a warnings ignore.

erykoff (changes requested):

python/lsst/meas/base/baseMeasurement.py
timj: or using `with self.assertWarns()`

parejkoj: I'll put in the warning, but if I recall correctly from Jim, there was no obvious path forward from here; the deprecation warning was mostly a "don't start using this" warning.

erykoff: Well, it should be warning of some sort, and we can catch that in the tests.  That way anybody else who tries this funny business can't say they weren't warned.

parejkoj: Looks like we have to ignore/catch them in the tests: when I ran scons, the tests all raised!

timj: `assertWarns` is going to complain if the warning does not appear so are there cases that do issue warnings and cases that don't?

erykoff: What do you mean?

timj: Are you asking me or John? I interpreted John's "the tests all raised" as meaning that some tests did issue warnings that the assertWarns caught but other tested wrapped in assertWarns did not issue warnings and so triggered test failures. If you aren't really sure which bit of code is going to warn or not then globally ignoring warnings for the test is the only choice (or else use assertWarns and then issue a warning of your own for it to catch).

erykoff: I was asking John, but I think you have the correct interpretation.  But I don't know why some tests wouldn't be warning.  I'll leave this ti John.

parejkoj: My point is that this `FutureWarning` raised and failed the tests. I wasn't aware that we'd switched this package to fail on warnings.

timj: As far as I'm aware no package is configured to fail tests if a warning is issued (all our packages would fail). Can you please show the output you get when your test fails (since `assertWarns` isn't even in play).

timj: `main` of `meas_base` issues 3 FutureWarning message already:

```
tests/test_MeasureSources.py::MeasureSourcesTestCase::testPeakLikelihoodFlux
  /Users/timj/work/lsstsw/build/meas_base/tests/test_MeasureSources.py:113: FutureWarning: Default position argument overload is deprecated and will be removed in version 24.0.  Please explicitly specify a position.
    psfKernel = psf.getLocalKernel()

tests/test_MeasureSources.py::MeasureSourcesTestCase::testPeakLikelihoodFlux
  /Users/timj/work/lsstsw/build/meas_base/tests/test_MeasureSources.py:114: FutureWarning: Default position argument overload is deprecated and will be removed in version 24.0.  Please explicitly specify a position.
    psfImage = psf.computeKernelImage()

tests/test_diaCalculation.py::TestDiaCalcluation::testRunUnindexed
  /Users/timj/work/lsstsw/build/meas_base/tests/test_diaCalculation.py:213: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
    unindexedDiaSources = unindexedDiaSources.append(
```

parejkoj:
```____________________ PsfFluxTestCase.testSingleFramePlugin _____________________
[gw4] linux -- Python 3.10.5 /home/parejkoj/lsst/lsstsw/miniconda/envs/lsst-scipipe-4.1.0/bin/python3.10

self = <test_PsfFlux.PsfFluxTestCase testMethod=testSingleFramePlugin>

    def testSingleFramePlugin(self):
>       task = self.makeSingleFrameMeasurementTask("base_PsfFlux")

tests/test_PsfFlux.py:215:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
python/lsst/meas/base/tests.py:666: in makeSingleFrameMeasurementTask
    config = self.makeSingleFrameMeasurementConfig(plugin=plugin, dependencies=dependencies)
python/lsst/meas/base/tests.py:624: in makeSingleFrameMeasurementConfig
    config = SingleFrameMeasurementTask.ConfigClass(ignoreSlotPluginChecks=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'lsst.meas.base.sfm.SingleFrameMeasurementConfig'>
ignoreSlotPluginChecks = True, args = (), kwargs = {}
instance = lsst.meas.base.sfm.SingleFrameMeasurementConfig(slots={'centroid': 'base_SdssCentroid', 'shape': 'base_SdssShape', 'ps...am': 0.0}, 'test_NoiseReplacer': {'doMeasure': True}, 'test_LoggingPlugin': {'doMeasure': True}}}, loggingInterval=600)
msg = 'ignoreSlotPluginChecks is deprecated and should only be used in tests. No removal date has been set yet.'

    def __new__(cls, *args, ignoreSlotPluginChecks=False, **kwargs):
        instance = super().__new__(cls, *args, **kwargs)
        if ignoreSlotPluginChecks:
            msg = ("ignoreSlotPluginChecks is deprecated and should only be used in tests."
                   " No removal date has been set yet.")
>           raise FutureWarning(msg)
E           FutureWarning: ignoreSlotPluginChecks is deprecated and should only be used in tests. No removal date has been set yet.

python/lsst/meas/base/baseMeasurement.py:175: FutureWarning```

timj: It's supposed to be a `warnings.warn("Message", category=FutureWarning, stacklevel=2)` not a `raise`.

parejkoj: Oh, I need to do `warnings.warn(msg, FutureWarning)`. Sorry about the chatter.

timj: Yes, `stacklevel=2` is the polite option to add as well because it means that the caller gets told the line of theirs that caused the warning to trigger, rather than the line in meas_base.

parejkoj: If that's a thing we should do in warnings, please add it to the dev guide deprecations section, possibly [here in package deprecation](https://developer.lsst.io/stack/deprecating-interfaces.html#package-deprecation), but maybe we should have a paragraph in the python section above that about deprecating function call signatures too?

erykoff (approved):

PR #228, opened 2022-12-22
DM-37428: Use image XY0 for pixel center calculation

isullivan:


TallJimbo (approved):
Certainly looks plausible that it fixes the problem, so if you say it does, this looks good to me.

PR #233, opened 2023-01-26
DM-37532: use pybind wrappers, create only one shared lib

mwittgen:


erykoff (approved):
Looks good; I guess we decided we don't need stub files and should just fix our own code like meas_extensions_gaap?

PR #237, opened 2023-04-12
tickets/DM-37814: Fix bug in heavy footprint stripping

enourbakhsh:


fred3m:
forcedPhotCoadd is no longer in meas_base so this PR is invalid.

PR #239, opened 2023-04-18
DM-31924: new high-level interface for making IDs for catalogs and their rows

TallJimbo:


python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+        Notes
+        -----
+        This method cannot be used on IDs generated with no data ID.
+        """
+        bits = _IdGeneratorBits(dimension_packer, n_releases)
+
+        def unpack(record_id: int) -> tuple[int, DataCoordinate, int]:
+            rest, counter = divmod(record_id, bits.n_counters)
+            rest, packed_data_id = divmod(rest, bits.n_data_ids)
+            rest, release_id = divmod(rest, bits.n_data_ids)
```
parejkoj, line 406: Oh, cool, I didn't know about `divmod` in python. That's handy.

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+            if rest:
+                raise ValueError(f"Unexpected overall factor {rest} in record_id {record_id}.")
```
parejkoj, line 408: Maybe include the other unpacked things for debugging? e.g. `... in record_id {record_id}, after unpacking counter={counter}, packed_data_id={packed_data_id}, release_id={release_id}`

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+
+    @property
+    def data_id(self) -> DataCoordinate:
+        """The data ID that will be embedded in all generated IDs
+        (`DataCoordinate`)."""
+        return self._data_id
+
+    @property
+    def release_id(self) -> int:
+        """The release ID that will embedded in all generated IDs (`int`)."""
```
parejkoj, line 458: No implementation?

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+        # Docstring inherited.
+        return IdFactory.makeSource(self.catalog_id, self._bits.counter_bits)
+
+    def arange(self, *args, **kwargs) -> np.ndarray:
+        # Docstring inherited.
+        lower = super().arange(*args, **kwargs)
+        if np.any(lower >= self._bits.n_counters):
+            arg_terms = [repr(arg) for arg in args] + [
+                f"{k}={v!r}" for k, v in kwargs.items()
+            ]
```
parejkoj, line 479: Maybe you ran black on it, but I'd have left the above all on one line, I think.

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+            raise ValueError(
+                f"Integer range from numpy.arange({arg_terms}) has values that are not "
```
parejkoj, line 481: Maybe "Integer range ({lower}) ...", or something else to include the failing value?

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+
+@dataclasses.dataclass
+class _IdGeneratorBits:
+    """A private helper struct that manages the allocation of bits between the
+    packed data ID, the release ID, and a per-catalog counter.
+    """
+
+    packer: DimensionPacker
+    """Object that maps data IDs to integers
+    (`lsst.daf.butler.DimensionPacker`).
```
parejkoj, line 495: I'm a bit surprised that we have to duplicate types for dataclasses. Isn't that one of the points of the dataclass definition?

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+
+    n_counters: int = dataclasses.field(init=False)
+    """Number of contiguous counter values to reserve space for, starting from
+    zero (`int`)."""
+
+    def __post_init__(self) -> None:
+        self.n_data_ids = 1 << self.packer.maxBits
+        upper_bits = (self.n_releases - 1).bit_length() + self.packer.maxBits
+        self.counter_bits = IdFactory.computeReservedFromMaxBits(upper_bits)
+        self.n_counters = 1 << self.counter_bits
```
parejkoj, line 517: The class docstring should probably note these three attributes, even though this is intended to be private.

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+# No deprecation decorator here because the type this class holds is itself
+# deprecated, so it's only going to be called be code paths that will go away
```
parejkoj, line 521:
```suggestion
# deprecated, so it's only going to be called by code paths that will go away
```

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+# when the deprecation turns into a removal, and which already warn.
```
parejkoj, line 522: Another place to note the removal ticket.

tests/test_id_generator.py
```diff
@@ -0,0 +1,142 @@
...
+        for _ in range(5):
+            catalog.addNew()
+        array = id_gen.arange(1, 6)
+        self.assertEqual(list(catalog["id"]), list(array))
+        if unpacker is not None:
+            expected_data_id = id_gen.data_id
+            for i in range(5):
+                embedded_release_id, embedded_data_id, counter = unpacker(array[i])
+                self.assertEqual(counter, i + 1)
+                self.assertEqual(embedded_data_id, expected_data_id)
```
parejkoj, line 137: I may be missing something, but I feel like this test is circular; it checks that the id you get out is what you passed in, no? Maybe that's all that is necessary and I'm missing something?

parejkoj (commented):
How is the test coverage of the new id classes? I'm having trouble figuring out how much of the new code would be executed by the tests you added, and whether the tests are actually independent of the code itself.

You say "integers" throughout this: is that intentionally "number of bits" generic, or do you mean 32-bit integers?

I think this could use some narrative documentation on how to use it, probably in `doc/lsst.meas.base/config/`.

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+
+        This combines the packed data ID and release ID, but not the
+        counter.
+        """
+        return 0
+
+    def __str__(self) -> str:
+        """Return a human-readable representation of the data ID (or a note
+        about its absence) for use in log and error messages.
+        """
```
parejkoj, line 296: Is the dataId the only thing we want to include in these `str`, or should it say something about the number of bits, too?

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+        This method is deprecated along with the type it accepts; it's provided
+        only as a temporary helper to aid in the transition from
+        `lsst.obs.base.ExposureIdInfo` to `IdGenerator`.
+        """
+        return _ExposureIdInfoIdGenerator(exposure_id_info)
+
+    @property
+    def catalog_id(self) -> int:
+        """The integer identifier for the full catalog with this data ID, not
+        just one of its rows (`int`).
```
parejkoj, line 286: As I dig into the uses of this, I am surprised that `exposureIdInfo.expId` is replaced with `idGenerator.catalog_id`. Thinking about it, the latter makes sense (you're using it for catalog-y things, not exposure-y things), but it might surprise others when they go to look at it ("why doesn't the new packer have an exposure_id I can get?"). Probably just worth being clear about this change when you announce the merging of this ticket.

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+        unpacker = IdGenerator.unpacker_from_config(
+            config.id_generator, butler.registry.expandDataId(skymap="HSC"),
+        )
+        release_id, embedded_data_id, counter = unpacker(catalog[0]["id"])
+        assert embedded_data_id == given_data_id
+
+    This example is a bit contrived, as the ability to reconstruct the data ID
+    is really only useful when you don't have it already, such as when the
+    record ID is obtained from some further-processed version of the original
+    table (such as a SQL database), and in that context the right config to
```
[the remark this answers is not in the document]
TallJimbo: The example includes a butler because I figured it was the most common way to get the right config, but you don't _need_ a butler to do the unpacking.  Just this morning @natelust was in my office talking about an idea to make it possible to get tasks (and by extensions, configs) by reading pipeline files, which would be another nice code path.

Neither of those is an ideal path for science users, at least not without some convenience code layered on top (some e.g.  `unpack_object_id` function that knows what config dataset types or pipelines to go read).  But I think the right answer for science users will have to come down to some sort of convenience code.

python/lsst/meas/base/_id_generator.py
TallJimbo: I was envisioning `str` here as being really for logs and exceptions about the catalog, not logs and exceptions about the IdGenerator itself.  There are a lot of tasks whose `run` methods take an `idGenerator` and that effectively abstracts over whether there is a data ID at all.

python/lsst/meas/base/_id_generator.py
TallJimbo: Yeah, I vaguely knew about it before, but I discovered while working on this ticket that it's just fantastic for making this kind of unpacking more readable.

python/lsst/meas/base/_id_generator.py
TallJimbo: I think most of the ways to get to this error involve those being garbage, because you're trying to unpack something generated some other way, but sure, it can't hurt.

python/lsst/meas/base/_id_generator.py
TallJimbo: Good catch!

python/lsst/meas/base/_id_generator.py
TallJimbo: I tried to come with a way to do that that doesn't potentially print a huge array, but couldn't come up with one that didn't involve trying to reimplement `np.arange` arg-handling.

python/lsst/meas/base/_id_generator.py
TallJimbo: It's just like any other type annotation as far as Sphinx is concerned right now.  I cannot way for the promised documenteer upgrade that will fix this.

python/lsst/meas/base/_id_generator.py
TallJimbo: What do you mean?  I don't think of attribute docs as going in the class docstring at all; the class `Parameters` is for `__init__` arguments, and these are not.

tests/test_id_generator.py
TallJimbo: It checks that the packing and unpacking methods are consistent with each other, and I figured that since they're effectively each independent implementations of the same algorithm, it wasn't worth it to write a third implementation in the test code.

TallJimbo (commented):
> How is the test coverage of the new id classes? I'm having trouble figuring out how much of the new code would be executed by the tests you added, and whether the tests are actually independent of the code itself.

Coverage of the new file is 89%; what's missing is mostly just the deprecated things (I've gotten accustomed to using `coverage` to help me write tests, too).

> You say "integers" throughout this: is that intentionally "number of bits" generic, or do you mean 32-bit integers?

I specifically mean 64-bit signed integers for the full row ID, since that's what afw.table uses (FITS says we can't have uint64, and various project-level docs says 64 bits).

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+from lsst.obs.base import ExposureIdInfo
+from lsst.pex.config import Config, ConfigField, Field
+from lsst.pipe.base import Instrument
+from lsst.skymap.packers import SkyMapDimensionPacker
+
+DEFAULT_RELEASE_ID = 0
+"""Default release ID to embed in catalog IDs.
+
+This can be changed globally to avoid having to override individual task
+configs to set the release ID.
```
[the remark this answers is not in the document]
parejkoj: Fair enough: do you think it's worth a short RFC to describe your idea for this process, and so we can think about other places we'll have to intentionally bump version numbers for releases? At least to have more than you and me being aware of this setting.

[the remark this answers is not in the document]
parejkoj: Oh, and a terrible thought: do we need to worry about point releases for data releases? I hope not!

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+    """
+
+    packer = SkyMapDimensionPacker.make_config_field()
+
+    def _make_dimension_packer(self, data_id: DataCoordinate) -> DimensionPacker:
+        # Docstring inherited.
+        return self.packer.apply(data_id)
+
+
+class IdGenerator:
```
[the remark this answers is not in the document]
parejkoj: In that case, should there be a note about which methods are typically overridden?

python/lsst/meas/base/_id_generator.py
parejkoj: We discussed this point on one of the other PRs; using the id generator to access a stringified dataId seems like a bit of an abuse of the system. Maybe just note in the `IdGenerator` docs that we use it as such?

python/lsst/meas/base/_id_generator.py
parejkoj: Oh! In that case, just include `has {(lower >= self._bits.n_counters).sum()} values that are ...`.

python/lsst/meas/base/_id_generator.py
parejkoj: Sorry about that: I see them as `init=False` fields now, with docstrings, so that's fine.

python/lsst/meas/base/_id_generator.py
```diff
@@ -0,0 +1,548 @@
...
+
+        Returns
+        -------
+        packer : `lsst.daf.butler.DimensionPacker`
+            Object that packs data IDs into integers.
+        """
+        raise NotImplementedError("Method is abstract.")
+
+
+class ExposureIdGeneratorConfig(BaseIdGeneratorConfig):
```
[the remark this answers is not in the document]
parejkoj: Yeah, it's very verbose, but I think that distinction is important here.

doc/lsst.meas.base/generating_source_and_object_ids.rst
```diff
@@ -0,0 +1,26 @@
...
+This is achieved by first packing the :ref:`Butler data ID <lsst.daf.butler-dimensions_data_ids>` that identifies an in-memory table object into an integer, and then packing that with a per-table counter.
+The packing algorithms are reversible (they're not just a hash), and we always aim to fit the full row ID into a single 64-bit signed integer.
+In special cases, like LSST data releases, an identifier for the release may be packed in as well, making the row ID globally unique (for its type) across all releases.
+
+The `IdGenerator` class is the main entry point for generating these IDs.
+Usage starts with with defining an `lsst.pex.config` field for the type of data ID that identifies a table object:
+
+- `SkyMapIdGeneratorConfig` for ``{tract, patch}`` or ``{tract, patch, band}}`` data IDs (most object tables);
+- `DetectorVisitIdGeneratorConfig` for ``{visit, detector}`` data IDs (most source tables);
+- `DetectorExposureIdGeneratorConfig` for ``{exposure, detector}`` data IDs (rare, at least in the default LSST data model).
```
parejkoj, line 19: I think these all need the full namespace in order to get built and linked.

doc/lsst.meas.base/generating_source_and_object_ids.rst
```diff
@@ -0,0 +1,26 @@
...
+
+These have a `~BaseIdGeneratorConfig.make_field` method that can be used to define the config field with minimal boilerplate.
```
parejkoj, line 21: I think the full namespace is necessary here, too? And in the paragraph below.

parejkoj (commented):
Thanks for the added narrative doc; a few comments on formatting there.

python/lsst/meas/base/_id_generator.py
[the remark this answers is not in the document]
TallJimbo: I think I'll add a quick note to RFC-917 when adopting it, so (as you said) what's present now has a bit more visibility.  I've got too many other things that need my attention right now to spend any more time on fleshing this scheme out.

python/lsst/meas/base/_id_generator.py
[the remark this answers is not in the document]
TallJimbo: I don't think so in this.  I don't expect any new derived classes and don't really want to advertise subclasses as something any user might want to do.

doc/lsst.meas.base/generating_source_and_object_ids.rst
TallJimbo: (see [above](https://github.com/lsst/meas_base/pull/239#discussion_r1178160180))

doc/lsst.meas.base/generating_source_and_object_ids.rst
parejkoj: Oh, that's cool! I wasn't aware of that. I'll have to remember it for my own docs.

PR #240, opened 2023-04-27
DM-38911: Add Compensated Gaussian plugin and associated tests.

erykoff:


tests/test_CompensatedGaussianFlux.py
```diff
@@ -0,0 +1,330 @@
...
+        xx, yy = np.meshgrid(np.arange(arr.shape[0]), np.arange(arr.shape[1]))
+        # Compute the inner and outer normalized Gaussian weights.
+        inner = (1./(2.*np.pi*width**2.))*np.exp(-0.5*(((xx - x_cent)/width)**2. + ((yy - y_cent)/width)**2.))
+        outer = (1./(2.*np.pi*(t*width)**2.))*np.exp(-0.5*(((xx - x_cent)/(t*width))**2.
+                                                           + ((yy - y_cent)/(t*width))**2.))
+        weight = inner - outer
+
+        # Compute the weighted sum of the pixels.
+        flux = np.sum(weight*arr)
+        # And the normalization term, derived in XXX.
```
natelust, line 116: Did you intend to fill in these XXX fields?

tests/test_CompensatedGaussianFlux.py
```diff
@@ -0,0 +1,330 @@
...
+        flux *= 4.*np.pi*(width**2.)*(t**2. + 1)/(t**2. - 1)
+
+        # And compute the variance
+        variance = np.sum(weight*weight*var_arr)
+        variance /= np.sum(weight*weight)
+
+        # The variance normalization term, derived in XXX.
```
natelust, line 123: and here

tests/test_CompensatedGaussianFlux.py
```diff
@@ -0,0 +1,330 @@
...
+                self.assertLess(np.abs(fit_x[0]), 1e-3)
+                fit_y = np.polyfit(y_sub, fluxes/50000.0, 1)
+                self.assertLess(np.abs(fit_y[0]), 1e-3)
+
+    def testCompensatedGaussianPlugin(self):
+        """Test for correct instFlux given known position and shape.
+        """
+        # In the z-band, HSC images have a noise of about 40.0 ADU, and a background
+        # offset of ~ -0.6 ADU/pixel.  This determines our test levels.
+        for width in (2, 3, 5):
```
natelust, line 232: These numbers are occurring everywhere, would it be better to to the like somewhere in case they needed changed in tandem at some point in the future

tests/test_CompensatedGaussianFlux.py
```diff
@@ -0,0 +1,330 @@
...
+                record = catalog[0]
+                algorithm.measure(record, exposure)
+                inst_flux = record[f"base_CompensatedGaussianFlux_{width}_instFlux"]
+
+                for noise in (0.001, 0.01, 0.1):
+                    fluxes = np.zeros(nSamples)
+                    errs = np.zeros_like(fluxes)
+
+                    for repeat in range(nSamples):
+                        # By using ``repeat`` to seed the RNG, we get results which
```
natelust, line 302: This is basically my problem with Monte Carlo unit tests. What you have here is fine, if possibly not strictly Monte Carlo. It is a good compromise and test behavior sufficiently

natelust (approved):
Minor points

tests/test_CompensatedGaussianFlux.py
erykoff: I have just copied this format from all the other tests in meas_base.

PR #243, opened 2023-05-09
DM-37196: Update fields to reflect new APDB schema

parejkoj:


mrawls (approved):
Thanks for updating the time fields to be `MjdTai` explicitly. Good to go.

PR #244, opened 2023-05-11
DM-37058: Disable unnecessary measurements in CCD forced photometry

parejkoj:
Also added some trivial tests of the two tasks in forcedPhotCcd.py

### DIFF
--- a/python/lsst/meas/base/forcedPhotCcd.py
+++ b/python/lsst/meas/base/forcedPhotCcd.py
@@ -251,6 +251,9 @@ class ForcedPhotCcdConfig(pipeBase.PipelineTaskConfig,
     def setDefaults(self):
         # Docstring inherited.
         super().setDefaults()
+        # Footprints here will not be entirely correct, so don't try to make
+        # a biased correction for blended neighbors.
+        self.measurement.doReplaceWithNoise = False
         # Only run a minimal set of plugins, as these measurements are only
         # needed for PSF-like sources.
         self.measurement.plugins.names = ["base_PixelFlags",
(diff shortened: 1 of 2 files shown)


python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -252,10 +250,21 @@ class ForcedPhotCcdConfig(pipeBase.PipelineTaskConfig,
 
     def setDefaults(self):
         # Docstring inherited.
+        super().setDefaults()
+        # TODO: Why not this one?
```
erykoff, line 254: A bare TODO?  Maybe it needs a `noqa`?  😄

erykoff (approved):

python/lsst/meas/base/forcedPhotCcd.py
parejkoj: Ah, right: this was an open question. The original `ForcedPhotCcdConfig` here doesn't set `doReplaceWithNoise=False` but `ForcedPhotCcdFromDataFrameConfig` does. What's the rationale for them being different? I added it here because the "fromDataFrame" code is newer, and this seems like something we wouldn't want for forced photometry anyway. Thoughts?

TallJimbo: You need heavy footprints from a deblender for `replaceWithNoise=True`, and you can't get those from a dataframe at all.

You can't _really_ get them from an afw.table.SourceCatalog in CCD forces phot, either, because the ones you've got are for the wrong coordinate system and PSF.  But I think there was some code path that downgraded the heavy footprints to regular ones as it transformed them, and I think if given those `replaceWithNoise=True` will just use them to replace other parent footprints with noise, not blended neighbors, which is of course much less useful.

So this is a story of the afw.table version making sort of a feeble attempt at handling blending because it was better than nothing (maybe?), and then this version not being able to do even that much.

Really handling blends here is really hard, which is why we'd really rather do all forced phot on diffims.

PR #247, opened 2023-06-23
DM-38546: Improve TestDataset for new CalibrateImage

parejkoj:


python/lsst/meas/base/tests.py
```diff
@@ -419,6 +419,11 @@ def _installFootprint(self, record, image, setPeakSignificance=True):
     def addSource(self, instFlux, centroid, shape=None, setPeakSignificance=True):
         """Add a source to the simulation.
 
+        To insert a point source with a given signal-to-noise (sn), the total
+        ``instFlux`` should be: ``sn*noise*psf_area``, where ``noise`` is the
+        noise you will pass to ``realize()``, and
+        ``psf_area=sqrt(4*pi*r^2)``, where ``r`` is the width of the PSF.
```
TallJimbo, line 434: I think the quantity inside the `sqrt` is more like an area; maybe `psf_size` instead of `psf_area` would be clearer.

TallJimbo (approved):

PR #257, opened 2023-10-27
DM-41345: Make TestDataset catalog look more like Calibrate output

parejkoj:


cmsaunders (approved):
This looks reasonable to me. I don't any of the other tests use the coordinates, so I guess moving `updateSourceCoords` into the lower-level class doesn't change anything.

PR #260, opened 2023-11-22
DM-37386: Add a CenterAll flag during detect and measure

parejkoj:


src/PixelFlags.cc
```diff
@@ -94,22 +117,46 @@ PixelFlagsAlgorithm::PixelFlagsAlgorithm(Control const& ctrl, std::string const&
...
-                                                               "Source's center is close to suspect pixels");
+    // Flags that correspond to mask bits which are set anywhere in the 3x3 central region of the object.
+    _centerKeys["INTRP"] = schema.addField<afw::table::Flag>(
+            name + "_flag_interpolatedCenter", "Interpolated pixel in the 3x3 region around the centroid.");
+    _centerKeys["SAT"] = schema.addField<afw::table::Flag>(
+            name + "_flag_saturatedCenter", "Saturated pixel in the 3x3 region around the centroid.");
+    _centerKeys["CR"] = schema.addField<afw::table::Flag>(
+            name + "_flag_crCenter", "Cosmic ray in the 3x3 region around the centroid.");
+    _centerKeys["BAD"] = schema.addField<afw::table::Flag>(name + "_flag_badCenter",
+                                                           "Bad pixel in the 3x3 region around the centroid");
```
fred3m, line 128: I'm not sure why this is formatted differently than all of the other schema fields.

fred3m (approved):
Ok, it took me a little while but I think I understand how this works now. I think that the use of `_bits` and `_bitsAll` is confusing, since `_bitsAll` doesn't really mean all of the bits in the footprint had a given masked bit set, just the center 3x3 grid, and when paired with `_bits` it's more confusing.

I understand what you were trying to convey, since `_bits` is the union of all of the bits in the spanset and `_bitsAll` is the intersection of all of the bits in the spanset, so I would recommend renaming them accordingly, ie. `_unionBits` and `_intersectionBits`, or `_anyBit` and `_allBits`.

src/PixelFlags.cc
parejkoj: Formatting is entirely determined by clang-format: in this case, I think it's because the docstring is shorter.

PR #264, opened 2024-02-15
DM-42914: Add edgeCenter and edgeCenterAll flags

parejkoj:


src/PixelFlags.cc
```diff
@@ -107,7 +107,8 @@ PixelFlagsAlgorithm::PixelFlagsAlgorithm(Control const& ctrl, std::string const&
     // Set all the flags that correspond to mask planes anywhere in the footprint
     _anyKeys["EDGE"] = schema.addField<afw::table::Flag>(
             name + "_flag_edge",
-            "Source is outside usable exposure region (masked EDGE or NO_DATA, or centroid off image)");
+            "Source is outside usable exposure region (masked EDGE or NO_DATA, or centroid off image).");
+    _anyKeys["NO_DATA"] = _anyKeys.at("EDGE");  // Also set edge flag for NO_DATA.
```
kfindeisen, line 111: Do I understand correctly that `_anyKeys[K]`, `_centerKeys[K]`, and `_centerAllKeys[K]` are "the key to the catalog flag that should be set if make K is set" (some comments in the class definition would have been helpful)? Should I be worried about the fact that the [config code](https://github.com/lsst/meas_base/blob/04a9dd8c0aad1c3faaccc75ddf2679f927b2838e/src/PixelFlags.cc#L156-L176) can apparently overwrite these keys?

parejkoj: Yes, your understanding of those maps is correct.

Good catch that we've never enforced that the config not change the existing key definitions. That should be fixed, but probably not on this ticket.

I can add some comments to the class header on this ticket: this code goes back to circa 2013 (thought my addition of `_centerAllKeys` is obviously newer).

src/PixelFlags.cc
```diff
@@ -219,15 +229,6 @@ void PixelFlagsAlgorithm::measure(afw::table::SourceRecord& measRecord,
...
 
-    // Set the EDGE flag if the bitmask has NO_DATA set
-    try {
-        if (func.getAnyBits() & MaskedImageF::Mask::getPlaneBitMask("NO_DATA")) {
-            measRecord.set(_anyKeys.at("EDGE"), true);
-        }
-    } catch (pex::exceptions::InvalidParameterError& err) {
-        throw LSST_EXCEPT(FatalAlgorithmError, err.what());
-    }
-
```
kfindeisen, line 230: Is this related to `_anyKeys["NO_DATA"] = _anyKeys.at("EDGE");`? If so, it should probably have been a separate commit (i.e., "clean up NO_DATA handling" as opposed to "Add edgeCenter*").

tests/test_PixelFlags.py
```diff
@@ -103,28 +108,50 @@ def testOffimageNonfinite(self):
...
         """
         # These four will be explicitly set to values at the boundary; we
         # cannot add sources off the image in TestDataset.
         self.dataset.addSource(100000, lsst.geom.Point2D(20, 20))
         self.dataset.addSource(100000, lsst.geom.Point2D(20, 100))
-        self.dataset.addSource(100000, lsst.geom.Point2D(80, 100))
+        self.dataset.addSource(100000, lsst.geom.Point2D(40, 100))
         self.dataset.addSource(100000, lsst.geom.Point2D(80, 30))
+        # Half on the edge; no offimage or edgeCenterAll, but edge and edgeCenter.
+        self.dataset.addSource(100000, lsst.geom.Point2D(115, 30))
```
kfindeisen, line 127: Does "half on the edge" mean the source sits on the border of the edge region? Is the centroid _exactly_ on the border, or merely <1px off (I can see different actual/desired behavior for the two cases)?

tests/test_PixelFlags.py
```diff
@@ -103,28 +108,50 @@ def testOffimageNonfinite(self):
...
         np.testing.assert_array_equal(catalog["base_PixelFlags_flag_offimage"],
-                                      np.array([False, False, True, False, True]))
+                                      np.array([False, False, True, False, True, False, False]))
         np.testing.assert_array_equal(catalog["base_PixelFlags_flag_edge"],
-                                      np.array([False, False, True, False, True]))
+                                      np.array([False, False, True, False, True, True, True]))
+        np.testing.assert_array_equal(catalog["base_PixelFlags_flag_edgeCenter"],
+                                      np.array([False, True, True, True, True, True, True]))
+        np.testing.assert_array_equal(catalog["base_PixelFlags_flag_edgeCenterAll"],
+                                      np.array([False, True, True, True, True, False, True]))
```
kfindeisen, line 154: This style of testing is very hard to validate (which index is which source? 😵‍💫), but I suppose that's out of scope for this issue.

kfindeisen (approved):
Looks good, just some clarification/style comments.

src/PixelFlags.cc
parejkoj: Fair point: I've made that a separate commit.

tests/test_PixelFlags.py
parejkoj: I've clarified the comments. It's about whether `edgeCenter` vs. `edgeCenterAll` is set. All that I care about is where central 3x3 region is, not the exact centroid.

kfindeisen: Yes, but the position of the 3×3 region will depend on where the centroid is, and by how much...

tests/test_PixelFlags.py
parejkoj: Yeah, I know. I am responsible for adding these tests myself, so I can't blame history on this part. It was useful to have them all in one test so I could look at them on the image with afw.display to check that I'd put them where I thought I had, and then to look at the arrays being tested to see that they covered the cases I wanted.

Suggestions on how to clean it up? I could pull each source into its own test, but then it's a bit harder to reason about having caught an appropriate number of cases.

kfindeisen: Yeah, a separate method for each would probably be best. The "loop over examples" idiom wouldn't really work here, especially if you need explanatory comments for each source. 🙂

include/lsst/meas/base/PixelFlags.h
```diff
@@ -78,9 +78,10 @@ class PixelFlagsAlgorithm : public SimpleAlgorithm {
 
 private:
     Control _ctrl;
-    KeyMap _centerKeys;
-    KeyMap _centerAllKeys;
-    KeyMap _anyKeys;
+    // The afw Table Keys for the various fields we add to the schema on construction.
+    KeyMap _centerKeys;     // Keys for bits set anywhere within the 3x3 region around the centroid.
+    KeyMap _centerAllKeys;  // Keys for bits set on the entire 3x3 region around the centroid.
+    KeyMap _anyKeys;        // Keys for bits set anywhere in the footprint.
```
kfindeisen, line 84: I think the comment should at least say that the key is the name of the mask plane.

parejkoj: Better?

    // The mask name->afw Table Key map for the various fields we add to the schema on construction.

tests/test_PixelFlags.py
parejkoj: I separated the edge from the offImage tests, so at least there are many fewer bools in the arrays. I think because they're grouped together and only 2 or 4 entries to check, this is clear enough now. Thoughts?

tests/test_PixelFlags.py
parejkoj: Do the updated comments help?

kfindeisen: I guess? I'm not sure how the pixels can be "half" on the edge (a pixel is either flagged or not, no?)

parejkoj: Ah, I meant "half of the" - I'll fix that.

kfindeisen: I meant that I'm not sure how you can get half of them like that -- I would have expected e.g. one row on and two off, or vice versa.

parejkoj: Pedantism strikes again! You're right, but that doesn't matter for the test itself (nothing cares about the centroid itself, just the box around it).

PR #266, opened 2024-03-11
DM-42811: Log unexpected measurement exceptions as WARNING

parejkoj:


TallJimbo (approved):
Could you add the context we discussed on slack to the commit message, i.e. "these were turned from WARNING to DEBUG on ticket X, which was _supposed_ to be about adding new DEBUG logs for other errors but communicated poorly".

PR #268, opened 2024-04-01
DM-41700: Deprecate Naive Centroid

parejkoj:


include/lsst/meas/base/NaiveCentroid.h
```diff
@@ -41,7 +41,9 @@ namespace base {
 /**
  *  @brief A C++ control class to handle NaiveCentroidAlgorithm's configuration
  */
-class NaiveCentroidControl {
+class [[deprecated(
+        "This algorithm and its control class are deprecated and will be removed after "
+        "v27.")]] NaiveCentroidControl {
```
TallJimbo, line 46: It'd be best to suppress these warnings in the pybind11 module so we don't see them when compiling that also-deprecated code.  The preprocessor directives to do that can be found in the [dev guide just above your new paragraph](https://developer.lsst.io/stack/deprecating-interfaces.html#c-deprecation).

tests/sillyCentroid.py
```diff
@@ -23,6 +23,6 @@
 from _sillyCentroid import SillyCentroidAlgorithm, SillyCentroidControl, SillyTransform
 
 # Do not register SillyCentroid in plugins.py, as it's not part of meas_base
-wrapSimpleAlgorithm(SillyCentroidAlgorithm, Control=SillyCentroidControl,
+wrapSimpleAlgorithm(SillyCentroidAlgorithm, Control=SillyCentroidControl, name="testLib_SillyCentroid",
```
TallJimbo, line 26: I think this algorithm is just on the border of being simple enough that it might have saved time to reimplement it in pure Python.  But now that you've done it this way that's probably not worth it.

TallJimbo (approved):

include/lsst/meas/base/NaiveCentroid.h
parejkoj: Thanks. I had to put the pragma around the entire pybind11 block in the namespace, but that did take care of the compiler warnings.

tests/sillyCentroid.py
parejkoj: Yeah, it took me a bit to sort out the various "testLib" references (and lack thereof), but this works fine. Having someone implement SillyCentroid (which is basically identical to NaiveCentroid) in python might be a useful training exercise for a newer developer, but is otherwise not worth it.

PR #270, opened 2024-04-18
DM-43969: Call MeasurementError correctly in sizeExtendedness plugin

arunkannawadi:


tests/test_MomentsClassifier.py
```diff
@@ -56,6 +56,20 @@ def testSingleFramePlugin(self):
         for ii in range(self.n_stars, self.n_stars + self.n_gals):
             self.assertGreater(catalog[ii].get("base_ClassificationSizeExtendedness_value"), 0.02)
 
+    def testFailure(self):
+        """Test that MeasurementError is raised properly."""
```
parejkoj, line 60: Please say here why MeasurementError should be raised in this test?

tests/test_MomentsClassifier.py
```diff
@@ -56,6 +56,20 @@ def testSingleFramePlugin(self):
...
+        config = measBase.SingleFrameMeasurementConfig()
+        task = self.makeSingleFrameMeasurementTask(config=config)
+        exposure, catalog = self.dataset.realize(10.0, task.schema, randomSeed=5)
+        plugin = task.plugins["base_ClassificationSizeExtendedness"]
+        plugin_order = plugin.getExecutionOrder()
+        task.run(catalog, exposure, endOrder=plugin_order)
+        catalog["slot_Shape_flag"] = 1
```
parejkoj, line 67: I assume this is what causes the raise?

tests/test_MomentsClassifier.py
```diff
@@ -56,6 +56,20 @@ def testSingleFramePlugin(self):
...
+        for record in catalog:
+            with self.subTest(id=record["id"]):
+                with self.assertRaises(measBase.MeasurementError):
```
parejkoj, line 70: Please use `assertRaisesRegex` to confirm that the error you expected is what is being raised. I think just "Shape flag is set." is enough.

parejkoj (approved):
A couple of clarification comments, otherwise this is good.

PR #271, opened 2024-04-19
DM-43982: Raise FatalAlgorithmError in sizeExtendedness plugin

arunkannawadi:


arunkannawadi (commented):
I had originally intended to catch this in `__init__` but we cannot guarantee that the required shape algorithms are registered before this. We should allow them to be registered in any order, since the plugins are ordered only at run time.

TallJimbo:
A measurement plugin's `__init__` should be called after the `__init__` of all of its dependencies, and if that's not happening it's a bug we can fix in the framework.  What I consider "registration" (adding the config entries to the `RegistryField`) does indeed happen out of order, but it should all be done before the first plugin `__init__` fires.

arunkannawadi:
You're absolutely right. The plugins are initialized according to their execution order value. So we can definitely move up this check during initialization which is still at runtime, but not during config validation time. A plugin's config knows nothing about other plugins.

python/lsst/meas/base/plugins.py
```diff
@@ -695,6 +701,16 @@ def getExecutionOrder(cls):
 
     def __init__(self, config, name, schema, metadata):
         SingleFramePlugin.__init__(self, config, name, schema, metadata)
+
+        # Check that the required columns are already in the schema.
+        if not all(("slot_Shape" in schema.getAliasMap().keys(),
+                    "slot_PsfShape" in schema.getAliasMap().keys(),
```
parejkoj, line 707: Do you have to get the alias map here, or can you just ask `"slot_Shape" in schema`? I just checked and that appears to do what you'd want.

python/lsst/meas/base/plugins.py
```diff
@@ -695,6 +701,16 @@ def getExecutionOrder(cls):
 
     def __init__(self, config, name, schema, metadata):
         SingleFramePlugin.__init__(self, config, name, schema, metadata)
+
+        # Check that the required columns are already in the schema.
+        if not all(("slot_Shape" in schema.getAliasMap().keys(),
```
parejkoj, line 706: Instead of `all`, just `and` them. There's on two things, no need for `all`.

python/lsst/meas/base/plugins.py
arunkannawadi: You'd have had to do `"slot_Shape_xx" in schema` and not what you typed, right? I could go either way, but I inclined towards the current implementation since it felt more generalizable without putting in the extra `_xx` by hand.

parejkoj:
Testing this against ap_verify, this doesn't actually solve the problem. It looks like I need to clear all the slots in CalibrateImageTask, as the schema there has the slots defined but they're not pointing to anything. For example `'slot_PsfShape'->'base_SdssShape_psf'`, but the latter doesn't exist (hence the error message in the ticket). I think you need to check both that the Shape slots have been defined, and that the thing they point to resolves. I'm not sure the best way to do that.

I'll separately fix CalibrateImage to remove all the nonexistent slots (probably om DM-43973).

arunkannawadi:
I can confirm the same. The previous version had `slots` pointing to something even though those plugins (`base_SdssShape_psf` in this case) were not being run at all. As we all learnt recently, thanks to John and Jim, it's just pointing to a string to substitute. I modified to look for the actual field in the schema the slots are pointing to, and `ap_verify` fails immediately as it should. The actual fix to avoid the failure is in DM-43973, so that blocks this now.

parejkoj (approved):
Thanks, I think this does take care of all the necessary cases.

PR #273, opened 2024-05-02
DM-44169: Fix compensated gaussian measurement error raising code

erykoff:


laurenam:
It seems you should refactor your commits...I don't think leaving the first one is a good idea (i.e. you now know `MeasurementError` should not be used, so history would prefer to forget!)  Also, the last commit now combines the code changes with the unittest updates.  I'd be inclined to keep them as separate commits (and just using the final form here).

erykoff:
Refactored the commits!

python/lsst/meas/base/compensatedGaussian/_compensatedGaussian.py
```diff
@@ -119,44 +105,68 @@ def __init__(
             mask_str = f"{base_key}_mask_bits"
             mask_key = schema.addField(mask_str, type=np.int32, doc="Mask bits set within aperture.")
 
-            self.width_keys[width] = (flux_key, err_key, mask_key)
+            # failure flags
+            failure_flag = flagDefs.add(f"{width}_flag", "Compensated Gaussian measurement failed")
+            oob_flag = flagDefs.add(f"{width}_flag_bounds", "Compensated Gaussian out-of-bounds")
+
```
laurenam, line 111: I think `width` evaluates to to a number (integer).  Do you think it will be apparent "enough" that this number in the flag name is the kernel width?  And will the casual user be expected to translate `bounds` to `outOfBounds` internally (probably yes to both, but my ~10pm brain didn't jump there right away)?

python/lsst/meas/base/compensatedGaussian/_compensatedGaussian.py
```diff
@@ -119,44 +105,68 @@ def __init__(
...
+                self.flagHandler.setValue(measRecord, oob_flag.number, True)
+                continue
+
             y_slice = slice(y_floor - rad, y_floor + rad + 1, 1)
             x_slice = slice(x_floor - rad, x_floor + rad + 1, 1)
             y_mean = y - y_floor + rad
             x_mean = x - x_floor + rad
 
+            sub_im = exposure.image.array[y_slice, x_slice]
+            sub_var = exposure.variance.array[y_slice, x_slice]
```
laurenam, line 154: You could technically move setting `sub_var` to after the size checking (but readability may win out to leaving here...up to you).

laurenam (approved):

python/lsst/meas/base/compensatedGaussian/_compensatedGaussian.py
erykoff: That's the name of the measurement ... `base_compensatedGaussian_5` or whatever.  It's what we decided on ages ago, and it may not even matter since the compensated gaussians are not working well as a calibration flux measurement.  😞 
As for the `bounds`, this is also the name from before, except done correctly.  I don't think this is going to be used for anything anyway.

PR #275, opened 2024-05-08
DM-44246: Add functionality to use an external footprint set for NoiseReplacer in SingleFrameMeasurementTask.

erykoff:


python/lsst/meas/base/baseMeasurement.py
```diff
@@ -485,3 +485,21 @@ def doMeasurementN(self, plugin, measCat, *args, **kwds):
                 self.log.getChild(plugin.name).warning(
                     "Exception in %s.measureN on records %s-%s: %s",
                     plugin.name, measCat[0].getId(), measCat[-1].getId(), error)
+
+    @staticmethod
+    def getFootprintsFromCatalog(catalog):
```
fred3m, line 490: I wonder if this should be in AFW instead. For example, it could be added to https://github.com/lsst/afw/blob/main/python/lsst/afw/table/_source.py#L33

tests/test_NoiseReplacer.py
```diff
@@ -104,6 +104,29 @@ def testNoiseReplacer(self, noiseSource, variance):
...
+        # We choose a random seed which causes the test to pass.
+        task = self.makeSingleFrameMeasurementTask("test_NoiseReplacer")
+        task.config.noiseReplacer.noiseSource = noiseSource
+        exposure, catalog = self.dataset.realize(variance, task.schema, randomSeed=0)
+        if noiseSource == 'meta':
+            md = PropertyList()
+            md['BGMEAN'] = variance
+            exposure.setMetadata(md)
+        footprints = task.getFootprintsFromCatalog(catalog)
+        task.run(catalog, exposure, footprints=footprints)
```
fred3m, line 119: Should you also check that the default (`footprints=None`) works?

fred3m (commented):
Just a couple of small comments. Mostly, I'm thinking that it would be useful to make `getFootprintsFromCatalog` a method of a `SourceCatalog`, since it would be useful to other users. In fact, I thought that it already was.

python/lsst/meas/base/baseMeasurement.py
erykoff: I thought about that briefly, but I think this particular dict format is specific for the noise replacer code used here, and thus should be in `meas_base`.  So maybe this should be a static method of `NoiseReplacer` instead?  (Though this would then make the call in `finalizeCharacterization` less obvious).

tests/test_NoiseReplacer.py
erykoff: That's already tested in the existing test that this is a copy of except pulling out the footprints.

python/lsst/meas/base/baseMeasurement.py
fred3m: Yeah, that's true. I think it's fine to leave it here.

PR #276, opened 2024-05-09
DM-44267: Add dummy wcs to ForcedPhotCcdFromDataFrameTask.

erykoff:


python/lsst/meas/base/forcedPhotCcd.py
```diff
@@ -550,8 +550,14 @@ def __init__(self, refSchema=None, initInputs=None, **kwargs):
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
 
-        # When run with dataframes, we do not need a reference wcs.
-        inputs['refWcs'] = None
+        # When run with dataframes, we do not need a real wcs.
```
TallJimbo, line 553: I think this statement is not quite right (I know it's preexisting, but it's also what I presume led you to assume a dummy WCS would work).  A true statement would that we do not need a real WCS as long as we only run plugins that only require a centroid from the reference catalog, not a shape: no CModel, no Kron, probably only certain GaAP fluxes.  And while I don't see a good reason to run any of those on DIAObjects, and that's the context in which we run this task in our pipelines, it's not hard to imagine someone trying to run it in some other context in which something like CModel or Kron might make sense.  And then those will blindly use the dummy PSF and spit out garbage.

So I think it'd be a bit better to add a `skyMap` connection and use `TractInfo.getWcs()`, since that actually should be the right WCS.

TallJimbo (approved):

PR #287, opened 2024-10-27
DM-47168: Catch calibrations set to None in measurement plugins that require them

laurenam:


python/lsst/meas/base/plugins.py
```diff
@@ -439,12 +439,14 @@ def __init__(self, config, name, schema, metadata):
                 "calibration factor at the location of the src.")
 
     def measure(self, measRecord, exposure, center):
-
         photoCalib = exposure.getPhotoCalib()
-        calib = photoCalib.getLocalCalibration(center)
+        if photoCalib is None:
+            calib = np.nan
+            calibErr = np.nan
```
erykoff, line 445: 1) Is it worth having a debug log here?  Could go either way.
2) There is also a field `base_LocalPhotoCalib_flag` that was set when this failed (and similarly `base_LocalWcs_flag`). I think we need to set these, but I'm not sure off the top of my head how to get to these field names ...

A-ha!  https://github.com/lsst/meas_base/blob/2f1216ac18b67d915507f5ef145f1b8649c3ee01/python/lsst/meas/base/wrappers.py#L547 shows that you can do `measRecord.set(self._failKey, True)` because this is a "generic plugin".

erykoff (changes requested):

tests/test_EvaluateLocalCalibration.py
```diff
@@ -53,6 +53,16 @@ def testPhotoCalib(self):
         self.assertEqual(record.get("base_LocalPhotoCalib"), calib)
         self.assertEqual(record.get("base_LocalPhotoCalibErr"), calibErr)
 
+    def testPhotoCalibIsNone(self):
+        task = self.makeSingleFrameMeasurementTask("base_LocalPhotoCalib")
+        exposure, catalog = self.dataset.realize(10.0, task.schema, randomSeed=0)
+        exposure.setPhotoCalib(None)
+        task.run(catalog, exposure)
```
erykoff, line 60: To confirm that things work as expected, you should also wrap this in a log checker, e.g.:
```
import logging  # at top

with self.assertNoLogs(level=logging.WARNING):
    task.run(catalog, exposure)
```

erykoff (approved):
Why did these flake8 problems come around now?

laurenam:
Just `cuz I tend to run `flake8` locally and it drives me bonkers when there are pre-existing (but apparently not checked in the repo?) errors unrelated to my changes...I can remove the commit if you don't think it belongs here.

PR #296, opened 2025-02-06
DM-47738: Integrate with new scarlet API

fred3m:


python/lsst/meas/base/forcedMeasurement.py
```diff
@@ -354,7 +354,14 @@ def run(self, measCat, exposure, refCat, refWcs, exposureId=None, beginOrder=Non
 
         # Create parent cat which slices both the refCat and measCat (sources)
         # first, get the reference and source records which have no parent
-        refParentCat, measParentCat = refCat.getChildren(0, measCat)
+        if 'deblend_nChild' in measCat.schema:
+            # Scarlet can create hierarcical blends, so we need a different
```
taranu, line 358: hierarchical.. but anyway, is everything else here going to work as expected including both parents and I guess grandparents?

fred3m: Yes, because the children are set by the `getChildren` method, which returns the children for a given parent. That means that for depth=1 sources that also have children, the depth=2 children have no knowledge of their grandparent other than being able to trace it using `catalog.find(catalog.find(child["parent"])["parent"])`, for example.

fred3m: The NaN values that I was getting a few weeks ago were because I didn't include this extra check to add the depth=1 parents to the list of parents. So existing testing did verify that this works now.

fred3m:
After a discussion with @TallJimbo we are simplifying the persistence model so that the intermediate parents do not have records in the source catalog, removing the need for this PR.

PR #297, opened 2025-02-08
DM-45545: Trap NaN before calling boost API

timj:
From v1.86 boost asserts that the bessel API is called with a positive value. Previously NaN was accepted. Now trap for NaN and return it before calling boost.

### DIFF
--- a/src/SincCoeffs.cc
+++ b/src/SincCoeffs.cc
@@ -39,7 +39,15 @@ namespace base {
 namespace {
 
 // Convenient wrapper for a Bessel function
-inline double J1(double const x) { return boost::math::cyl_bessel_j(1, x); }
+inline double J1(double const x) {
+    // Some versions of boost assert that x >= 0
+    // which fails for NaN. Retain previous behavior
+    // and return NaN for NaN.
+    if (std::isnan(x)) {
+        return x;
+    }
+    return boost::math::cyl_bessel_j(1, x);
+}
 
 // sinc function
 template <typename T>


mwittgen:
cyl_bessel_j is defined since C++17 in the standard library.
"If the argument is NaN, NaN is returned and domain error is not reported."
https://en.cppreference.com/w/cpp/numeric/special_functions/cyl_bessel_j

src/SincCoeffs.cc
```diff
@@ -39,7 +39,13 @@ namespace base {
 namespace {
 
 // Convenient wrapper for a Bessel function
-inline double J1(double const x) { return boost::math::cyl_bessel_j(1, x); }
+inline double J1(double const x) {
```
mwittgen, line 42: `const` is redundant if passed by value

src/SincCoeffs.cc
```diff
@@ -39,7 +39,13 @@ namespace base {
...
+    // Some versions of boost assert that x >= 0.
```
mwittgen, line 43: Should be comment say return NaN if the argument is NaN?

mwittgen (approved):

timj:
@mwittgen you mean that the new boost added a a behavior that is incompatible with the standard???

src/SincCoeffs.cc
timj: I'm not going to get into changing things outside the scope of the fix at this time.

mwittgen:
Looks like it

PR #298, opened 2025-02-14
DM-48928: Fix serialization issue with LombScarglePeriodogramMulti plugin

jeffcarlin:
The solution I came up with is to create separate columns for each band rather than two columns populated with multi-element lists in each row.

I also updated the unit tests so that we are testing the multi-band version of Lomb-Scargle.

The plugin works now, but it is slow. I think this is simply because the Astropy Lomb-Scargle calculation is slow, and not a limitation of our pipelines code.

### DIFF
--- a/python/lsst/meas/base/compensatedGaussian/_compensatedTophat.py
+++ b/python/lsst/meas/base/compensatedGaussian/_compensatedTophat.py
@@ -44,7 +44,7 @@ class SingleFrameCompensatedTophatFluxConfig(SingleFramePluginConfig):
         doc="The aperture radii (in pixels) to measure the top-hats.",
         dtype=int,
         minLength=1,
-        default=[12,],
+        default=[12, ],
     )
     inner_scale = RangeField(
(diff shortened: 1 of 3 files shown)


python/lsst/meas/base/compensatedGaussian/_compensatedTophat.py
```diff
@@ -44,7 +44,7 @@ class SingleFrameCompensatedTophatFluxConfig(SingleFramePluginConfig):
         doc="The aperture radii (in pixels) to measure the top-hats.",
         dtype=int,
         minLength=1,
-        default=[12,],
+        default=[12, ],
```
ebellm, line 47: spurious change

ebellm (approved):

python/lsst/meas/base/compensatedGaussian/_compensatedTophat.py
jeffcarlin: That was needed so that `scons` would complete (linting issue).

PR #302, opened 2025-03-28
DM-44030: Remove deprecated NaiveCentroid

isullivan:
This removal was intended to be complete on DM-41701, but was missed at that time.


parejkoj (approved):
Thanks for catching the reference in the docs, too!

PR #303, opened 2025-03-31
DM-46908: Cleanups of SdssCentroid while investigating negative sources

parejkoj:


python/lsst/meas/base/tests.py
```diff
@@ -485,7 +498,7 @@ def addSource(self, instFlux, centroid, shape=None, setPeakSignificance=True):
         image = self.drawGaussian(self.exposure.getBBox(), instFlux,
                                   lsst.afw.geom.Ellipse(fullShape, centroid))
         # Generate a footprint for this source
-        self._installFootprint(record, image, setPeakSignificance)
+        self._installFootprint(record, image, setPeakSignificance, negative=negative)
```
nevencaplar, line 501: negative=negative? Just checking, is this what you intended?

nevencaplar (approved):
Looks good to me

python/lsst/meas/base/tests.py
parejkoj: Yes: the user calls `addSource` and supplies `negative` if desired, this just passes it through.

PR #304, opened 2025-04-09
DM-49997: Reserve space for 64 data release IDs by default.

TallJimbo:


python/lsst/meas/base/_id_generator.py
```diff
@@ -47,7 +47,7 @@
 configs to set the release ID.
 """
 
-DEFAULT_N_RELEASES = 1  # 1 means don't reserve space for releases.
+DEFAULT_N_RELEASES = 64
```
timj, line 50: I love the optimism of keeping at this until 2085 or so...

PR #308, opened 2025-05-19
DM-45569: Add a color extractor information.

PFLeget:


python/lsst/meas/base/__init__.py
```diff
@@ -46,3 +46,4 @@
 from .transforms import *
 from .wrappers import *
 from .compensatedGaussian import *
+from .colorUtilities import *
```
cmsaunders, line 49: It looks like these are arranged alphabetically, except for `.compensatedGaussian`. I would the new line in alphabetical order, and fix `.compensatedGaussian` too.

python/lsst/meas/base/colorUtilities.py
```diff
@@ -0,0 +1,43 @@
...
+import numpy as np
+from lsst.afw.image import Color
+
+__all__ = (
+    "colorExtractor",
+)
+
+
+def colorExtractor(record, columnNameValue="psf_color_value", columnNameType="psf_color_type"):
+    try:
```
cmsaunders, line 34: Are you sure you want this to fail silently? Do you need it to just return an empty color sometimes, or would it be better to raise an exception if the `columnNameValue` or `columnNameType` isn't in the record?

cmsaunders (approved):
I have a couple small comments. It's up to you to decide what is best to do for the try/except.

PR #311, opened 2025-07-14
DM-51687: Simplify single visit forced photometry

fred3m:


python/lsst/meas/base/applyApCorr.py
```diff
@@ -244,6 +219,15 @@ def run(self, catalog, apCorrMap):
         # Wrap the task logger to a periodic logger.
         periodicLog = PeriodicLogger(self.log)
 
+        if isinstance(catalog, astropy.table.Table):
+            # Prepare the astropy table for use with this task.
+            self._prepAstropyTable(catalog)
+            xColumn = "x"
+            yColumn = "y"
```
erykoff, line 226: I would prefer if you make the column names configurable.  I know of at least one place I could use this on a pre-transformed table.

python/lsst/meas/base/applyApCorr.py
```diff
@@ -253,63 +237,86 @@ def run(self, catalog, apCorrMap):
                 self.log.warning("Cannot aperture correct %s because could not find %s in apCorrMap",
                                  apCorrInfo.name, " or ".join(missingNames))
                 for source in catalog:
-                    source.set(apCorrInfo.apCorrFlagKey, True)
+                    source[apCorrInfo.apCorrFlagName] = True
```
erykoff, line 240: Oh this is horrifying.  This should not be looping over sources.  But we do need to be a bit clever because of the way that afw works (but this should work with either):
```
catalog[apCorrInfo.apCorrFlagName] = np.ones(len(catalog), dtype=np.bool_)
```
(This is the same pattern as is used below)

python/lsst/meas/base/baseMeasurement.py
```diff
@@ -137,7 +137,42 @@ def setupSchema(self, schema):
             aliases.set("slot_CalibFlux", self.calibFlux)
 
 
-class BaseMeasurementConfig(lsst.pex.config.Config):
+class SimpleBaseMeasurementConfig(lsst.pex.config.Config):
```
erykoff, line 140: Is the `Simple` deliberate because it has some connection to the `SimpleAlgorithm` super-class, or is this a coincidence?  If so a docstring clarifying that would be helpful.  (Or if not there's a confusing name clash here)

python/lsst/meas/base/baseMeasurement.py
```diff
@@ -433,6 +422,46 @@ def doMeasurement(self, plugin, measRecord, *args, **kwds):
                 plugin.name, measRecord.getId(), error)
             plugin.fail(measRecord)
 
+
+class BaseMeasurementTask(SimpleBaseMeasurementTask):
+    """Ultimate base class for all measurement tasks.
```
erykoff, line 427: This docstring is no longer true.  Could you also specify above what is `Simple` there and not simple here?

python/lsst/meas/base/baseMeasurement.py
```diff
@@ -433,6 +422,46 @@ def doMeasurement(self, plugin, measRecord, *args, **kwds):
...
+        run. If `None`, an empty `~lsst.daf.base.PropertyList` will be
+        created.
+    **kwds
+        Additional arguments passed to `lsst.pipe.base.Task.__init__`.
+
+    Notes
+    -----
+    This base class for `SingleFrameMeasurementTask` and
+    `ForcedMeasurementTask` mostly exists to share code between the two, and
+    generally should not be used directly.
```
erykoff, line 442: This note is repeated, and again I'm at a loss as to the difference between the two.

python/lsst/meas/base/simple_forced_measurement.py
```diff
@@ -0,0 +1,252 @@
...
+            to determine the footprint boundary.
+        """
+        psf = exposure.getPsf()
+        if psf is None:
+            raise RuntimeError("Cannot construct Footprints from PSF shape without a PSF.")
+        bbox = exposure.getBBox()
+        wcs = exposure.getWcs()
+        for record in sources:
+            # TODO: vectorize this WCS call
+            localPoint = wcs.skyToPixel(record.getCoord())
```
erykoff, line 241: `wcs.skyToPixelArray(sources["ra"], sources["dec"], degrees=False)` (assuming radians).

You can also vectorize the contains call, I did that on the previous ticket.  `inBBox = np.atleast_1d(lsst.geom.Box2D(bbox).contains(x, y))` or something like that.

You still have to loop to make the footprints, I assume.

erykoff (commented):
Do we have tests of the new simple footprint adding?

python/lsst/meas/base/applyApCorr.py
```diff
@@ -194,7 +169,7 @@ class ApplyApCorrTask(lsst.pipe.base.Task):
     ConfigClass = ApplyApCorrConfig
     _DefaultName = "applyApCorr"
 
-    def __init__(self, schema, **kwds):
+    def __init__(self, schema=None, **kwds):
```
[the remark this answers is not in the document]
fred3m: Yeah, that's the idea. I tested with and without (notice that I deprecated `ForcedPhotCcd` so it still works even though it is no longer used in the science pipelines with this ticket) and it works fine.

python/lsst/meas/base/baseMeasurement.py
fred3m: No. This is something that Jim and I pair coded together to avoid having to have a huge refactor by changing the meaning of `BaseMeasurementTask`. As the note in `SimpleBaseMeasurementTask` says, it primarily exists to allow code to be shared between `SimpleForcedMeasurementTask` and `ForcedMeasurementTask`, where `SimpleForcedMeasurementTask` is a simplified version of forced measurement that assumes that uses a PSF-based footprint for every source, eliminating the need to transform the footprints. So this is really the workhorse of this ticket.

python/lsst/meas/base/baseMeasurement.py
fred3m: I'm clarifying in the docstrings now.

python/lsst/meas/base/baseMeasurement.py
fred3m: Looking back at the code now I think the idea was that forced photometry on CCDs doesn't need all of the config options from `BaseMeasurementTask` so Jim just wanted to create a base task for the base task. If you have suggestions for a better naming convention I'm happy to hear it. We could also take it to `dm-naming-things`.

erykoff (approved):

PR #326, opened 2026-06-03
DM-54105: Compute DiaObject position uncertainty in MeanDiaPosition

isullivan:
Add raErr, decErr, and ra_dec_Cov outputs to the ap_meanPosition plugin. The uncertainty combines a formal term (inverse-variance combination of per-source centroid errors, with a 2x2 stack when ra_dec_Cov is present) and a scatter term (standard error of the mean from the tangent-plane spread of source positions).

### DIFF
--- a/python/lsst/meas/base/diaCalculationPlugins.py
+++ b/python/lsst/meas/base/diaCalculationPlugins.py
@@ -515,14 +515,35 @@ class MeanDiaPositionConfig(DiaObjectCalculationPluginConfig):
 
 @register("ap_meanPosition")
 class MeanDiaPosition(DiaObjectCalculationPlugin):
-    """Compute the mean position of a DiaObject given a set of DiaSources.
+    """Compute the mean position and position uncertainty of a DiaObject
+    from its associated DiaSources.
+
+    The reported (ra, dec) is the inverse-variance weighted spherical
+    mean of the per-source positions, using only the DiaSources whose
+    per-source coordinate errors are finite.  The reported covariance
+    is the covariance of that weighted-mean estimator,
+
+        C_formal = (sum_i inv(C_i))^-1,
+
+    where ``C_i`` is the per-source 2x2 covariance, multiplied by a
+    *chi-squared scale factor* that inflates the result when the
+    empirical scatter is larger than the per-source errors predict:
+
+        chi2  = sum_i r_i^T inv(C_i) r_i      (r_i = tangent-plane
+                                               residual from the mean)
+        dof   = 2 * (N - 1)
+        C_obj = max(1, chi2 / dof) * C_formal
+
+    When the data agree with the per-source errors (chi2 ~ dof) the
+    scale factor is unity and ``C_obj == C_formal``.  When the residuals
+    exceed what the errors predict, the covariance is inflated.
     """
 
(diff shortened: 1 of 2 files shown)


python/lsst/meas/base/diaCalculationPlugins.py
```diff
@@ -546,23 +554,118 @@ def calculate(self, diaObjects, diaSources, **kwargs):
...
+
         def _computeMeanPos(df):
+            # First check if the diaSource history is physical.
             coords = list(geom.SpherePoint(src["ra"], src["dec"], geom.degrees)
                           for idx, src in df.iterrows())
             aveCoord = geom.averageSpherePoint(coords)
+            maxSep = max(aveCoord.separation(coord).asArcseconds()
+                         for coord in coords)
+            if maxSep > maxAllowedSep:
+                raise UnphysicalDiaSourceSeparation(maxSep, maxAllowedSep)
```
fred3m, line 586: This confuses me a little. Why is it that there is a maxSep? Does the above contain *all* of the positions of the dia source or just two? If the latter then that's fine, but if it's a source that has been observed multiple times doesn't it make sense that the average position could be further than some of the earlier/later observations?

fred3m (changes requested):
I think that this algorithm is flawed for a few reasons:
1. You calculated two different types of variances/covariances. Your `varRaFormal`, `varDecFormal`, and `covFormal` are all covariances for a weighted mean. You calculate an unweighted mean in `aveCoord`, so you are underestimating the covariance and variance here unless all of the sources have equal variance.
2. Your varRaScatter, varDecScatter, and covScatter use the unweighted mean. This matches your actual mean, but when you add them together you are adding estimates of two different estimators, so this combination is invalid. The two estimators will not agree, even if the errors are correct.
3. When we spoke you acknowledge the double counting but I think it's worth noting that is happening and I don't *think* it's what you actually want.

My guess is that the intent was to have one measurement of variance that takes the errors into account (otherwise why are we calculating them) and the other is intended to base it off of the intrinsic scatter in case our errors are wrong. I don't think the way that you're attempting to do this is correct.

python/lsst/meas/base/diaCalculationPlugins.py
isullivan: The default sets this to 3x the default maximum association radius, so it should only trigger if there is a real bug. I could probably remove this check and the associated config, since it shouldn't ever trigger.

isullivan:
@fred3m Please take another look when you get a chance. I have replaced the "formal + scatter" variance with a term that inflates the formal variance when the scatter is larger than the per-diaSource errors would suggest.

python/lsst/meas/base/diaCalculationPlugins.py
ebellm: This config is a circuit breaker--we had a pretty catastrophic bug (due to pandas int-float roundtripping) in early operations where DIASources were returned as "associated" that were many arcminutes away--this leads the DIAObject positions to move by large amounts on sky and makes them unusable scientifically.

I am happy to have a very conservative value set for `maxSep` but I would not be in favor of removing it.

fred3m (approved):
LGTM

